PR #13, opened 2021-05-01
Fixed libobs case in FindLibObs.cmake

SingingSword:
Avoids this warning in cmake .. on fedora33:

CMake Warning (dev) at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:426 (message):
  The package name passed to `find_package_handle_standard_args` (Libobs)
  does not match the name of the calling package (LibObs).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.

### DIFF
--- a/external/FindLibObs.cmake
+++ b/external/FindLibObs.cmake
@@ -86,7 +86,7 @@ if(MSVC)
 endif()
 
 include(FindPackageHandleStandardArgs)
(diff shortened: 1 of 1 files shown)


royshil:
Are you able to test in osx and Windows?

It's ok if you can't. I think this is anyway a good change. 

I'll merge later.

Thanks for the contribution

PR #15, opened 2021-05-11
Update README.md

MatiMax:
I've updated the macOS install section as it is not recommended to tamper with the contents of the application bundle itself but instead using the resource folder in the system's Library Application Support folders.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,10 +64,17 @@ $ cpack
 ```
 
 #### Install
-Unpack the package directly to your OBS install directory, e.g.
+Unpack the package to the plugins directory of the system's Library folder (which is Apple's preffered way)
+```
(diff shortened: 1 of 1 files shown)


MatiMax:
The update would only apply if, and only if the binary was signed by the developer, and the folder layout of the now available installer ZIP file would follow this pattern:
```
/obs-backgroundremoval/bin/obs-backgroundremoval.so
/obs-backgroundremoval/data/locale/en-US.ini
/obs-backgroundremoval/data/modnet_simple.onnx
/obs-backgroundremoval/data/SINet_Softmax_simple.onnx
```

The content of such a ZIP file could then be copied to the system or user `Library` folder on macOS.

royshil (approved):
Thanks!

PR #30, opened 2021-07-13
Feature - Process Mask every X frames of video

iBenMoore:
Added the feature to Process the Mask every X frames of video.  This can greatly decrease the CPU usage of the plugin making it usable on lower-spec'd systems, and people may even prefer the visual effect of processing the mask less frequently as it can make the contours of the mask less "jittery".


royshil:
@iBenMoore i've created another PR based on your contribution (Thank you!) just fixing some small things

Issue #56, opened 2021-11-18
Linux (Ubuntu 21.10) GPU Instructions


mproffitt:
I haven't seen anything in the documentation on this, and the code is geared around disabling GPU on any platform other than Windows but I've been playing around with this plugin today and was able to get this working with CUDA on Ubuntu Studio 21.10

> Use this at your own risk - I'm not responsible if you break your card.

To enable this, you need to patch 3 files:

- src/Model.h
- src/background-filter.cp
- CMakeLists.txt

### Patch file
[gpu-linux.diff.txt](https://github.com/royshil/obs-backgroundremoval/files/7565326/gpu-linux.diff.txt)

Apply the patch file with `git apply gpu-linux.diff.txt`

### onnxruntime
Linking needs to be with the GPU version of onnx - you can find the 1.7.0 release of this here: https://github.com/microsoft/onnxruntime/releases/download/v1.7.0/onnxruntime-linux-x64-gpu-1.7.0.tgz

Alternatively the 1.9.0 release also seems to work and can be found here: https://github.com/microsoft/onnxruntime/releases/download/v1.9.0/onnxruntime-linux-x64-gpu-1.9.0.tgz

If you're using the GPU version of onnxruntime, you can't seem to switch between CPU and GPU - I haven't looked to see if this is possible, nor am I likely to try.

### Building
The build procedure is the same as already documented

```
mkdir build && cd build
cmake .. -DWITH_CUDA=ON && cmake --build . && sudo cmake --install .
```

### Install (Ubuntu bundled, not PPA):

```
mkdir -p ~/.config/obs-studio/plugins/obs-backgroundremoval/{bin,data}
cp /usr/local/lib/obs-plugins/obs-backgroundremoval.so ~/.config/obs-studio/plugins/obs-backgroundremoval/bin
cp -r /usr/local/share/obs/obs-plugins/obs-backgroundremoval/* ~/.config/obs-studio/plugins/obs-backgroundremoval/data
```

### Tested with NVIDIA RTX 3070 TI

The Ubuntu PPA for OBS Studio and bundled OBS Studio both seem to work fine although the bundled version takes a little more setting up.

It seems to run  on GPU, I've been running it for a few hours with a video background with the following observations:

- GPU warms up to about 58°C then the fan kicks in and brings it down
- GPU processing is low - around 9 or 10⅜ on my card
[...]

PR #57, opened 2021-11-19
Linux GPU Usage

mproffitt:
This commit patches for using the GPU on Linux as detailed in https://github.com/royshil/obs-backgroundremoval/issues/56

Additionally it expands on the installation instructions for Ubuntu.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,10 +38,10 @@ if(WIN32 OR MSVC)
     find_path(TIFF_INCLUDE_DIR tiff.h PATHS ${VCPKG_INSTALLED_DIR}/include)
 
     find_file(Protobuf_PROTOC_EXECUTABLE protoc.exe PATHS ${VCPKG_INSTALLED_DIR}/tools)
-else()
-    # Apple or linux
+elseif(APPLE)
+    # Apple
     if (WITH_CUDA)
-        message(FATAL "Cannot use CUDA on non-Windows builds")
+        message(FATAL "Cannot use CUDA on Apple builds")
     endif()
 endif()
 
--- a/README.md
+++ b/README.md
@@ -79,17 +79,48 @@ The first is recommended as it preserves the plugins over the parallel installat
 ### Linux
 
 #### Ubuntu
+
 ```
-$ apt install -y libobs-dev libopencv-dev language-pack-en wget git build-essential cmake
+$ sudo apt install -y libobs-dev libopencv-dev language-pack-en wget git build-essential cmake
 $ wget https://github.com/microsoft/onnxruntime/releases/download/v1.7.0/onnxruntime-linux-x64-1.7.0.tgz
-$ tar xzvf onnxruntime-linux-x64-1.7.0.tgz --strip-components=1 -C /usr/local/ --wildcards "*/include/*" "*/lib*/"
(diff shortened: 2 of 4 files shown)


lukasmichel:
The file `cuda_provider_factory.h` seems to not exist in the Linux GPU version of onnxruntime and thus building fails. The file seems to be only present in the Windows version. As far as  I understand it, this include is only necessary on Windows.
This fixes building for me:
```
diff --git a/src/Model.h b/src/Model.h
index 56e979e..a711539 100644
--- a/src/Model.h
+++ b/src/Model.h
@@ -6,15 +6,16 @@
 #include <onnxruntime/core/providers/cpu/cpu_provider_factory.h>
 #else
 #include <onnxruntime_cxx_api.h>
+#ifdef _WIN32
 #include <cpu_provider_factory.h>
 #endif
-
-#ifdef WITH_CUDA
-#include <cuda_provider_factory.h>
 #endif
+
 #ifdef _WIN32
-#ifndef WITH_CUDA
+#ifdef WITH_CUDA
 #include <dml_provider_factory.h>
+#else
+#include <cuda_provider_factory.h>
 #endif
 #include <wchar.h>
 #endif
diff --git a/src/background-filter.cpp b/src/background-filter.cpp
index 7228d54..b29035b 100644
--- a/src/background-filter.cpp
+++ b/src/background-filter.cpp
@@ -8,11 +8,11 @@
 #include <onnxruntime_cxx_api.h>
 #include <cpu_provider_factory.h>
 #endif
+
+#ifdef _WIN32
 #ifdef WITH_CUDA
 #include <cuda_provider_factory.h>
-#endif
-#ifdef _WIN32
-#ifndef WITH_CUDA
+#else
 #include <dml_provider_factory.h>
 #endif
 #include <wchar.h>
```

mproffitt:
Good spot on this - I had the file installed but it looks like that is removed in the 1.9.0 distribution of onxruntime *but is present* in the 1.7.0 release which is why I didn't notice this missing when I rebuilt with 1.9.0

lukasmichel:
That's unfortunate. This means either developing for an old version of onnxruntime or upgrading to 1.9.0 and removing this include also for the Windows version. 
I actually did something rather similar to what you did a few weeks ago using version 1.9.0. It works well for daily use and did not require additional changes, but I also was not able to make switching between CPU and GPU work, which is why I didn't do a PR. I wanted to attempt to solve this first.

mproffitt:
The first way I noticed that this was actually running on my GPU rather than the CPU is the CPU usage dropped dramatically - I'm running a brand new machine with a completely fresh (mostly unpolluted) version of Ubuntu Studio and monitoring the output from htop was showing CPU at around 400% until I got GPU working, then the CPU happily idles at around 5% for OBS.

I don't find the GPU usage high for this plugin but GPU memory does rise rapidly. Try running nvidia-smi with `watch -n 0.001 nvidia-smi` - when you activate the plugin, you should see OBS appear in the process list and ramp up memory usage to around 1.2GB with C+G in the Type field

imagen

ngoonee:
@royshil could we get this merged?

Mte90:
To get compiled on Linux I had to apply this patch https://github.com/royshil/obs-backgroundremoval/pull/57#issuecomment-974374277 with onxx 1.9.0.

PR #108, opened 2022-08-21
Updated to build against obs 27.2.3 on Ubuntu 22.04

directrix1:
Not sure if the CMake changes will work correctly on Windows, but this works on Linux. Tested CPU. I don't have access to CUDA.


directrix1:
I might add, I am very unfamiliar with CMake. So definitely review those changes.

royshil:
we now support OBS 28+ on Ubuntu from #152 
closing this PR

Issue #125, opened 2022-10-07
‘OrtSessionOptionsAppendExecutionProvider_DML’ was not declared in this scope


LuisGFer:
In Ubuntu 22.04 when I try to cmake --build . following the installation instructions I get this error:

error: ‘OrtSessionOptionsAppendExecutionProvider_DML’ was not declared in this scope; did you mean ‘OrtSessionOptionsAppendExecutionProvider_CPU’?
  215 |             Ort::ThrowOnError(OrtSessionOptionsAppendExecutionProvider_DML(sessionOptions, 0));
      |                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                               OrtSessionOptionsAppendExecutionProvider_CPU
gmake[2]: *** [CMakeFiles/obs-backgroundremoval.dir/build.make:76: CMakeFiles/obs-backgroundremoval.dir/src/background-filter.cpp.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:83: CMakeFiles/obs-backgroundremoval.dir/all] Error 2
gmake: *** [Makefile:156: all] Error 2

Some Idea to solve it? Thanks

bobek:
Changing the mentioned line to `OrtSessionOptionsAppendExecutionProvider_CPU` (mind the `CPU` at the end) will result in successful build, with ONNX using just CPU (in my case it will result in 3 fully utilized cores).

DML stand for DirectML -- https://onnxruntime.ai/docs/execution-providers/DirectML-ExecutionProvider.html , which gets selected if `CUDA` is not available, but DML is not available on Linux (I am on Debian/sid).

Thus I would propose having `CPU` executor as a fallback.

rigo:
Thanks @bobek I successfully compiled and used obs-backgroundremoval with OBS 28.0.1 on Ubuntu 22.10.

PR #126, opened 2022-10-19
Don't attempt to use DML backend on non-Windows

### LINKED ISSUE #114: Compilation on Linux+AMD broken - OrtSessionOptionsAppendExecutionProvider_DML
The changes to background-filter.cpp in commit cc9d4a5711f9388ed110230f9f793bb071577a23 broke compilation in Linux (Ubuntu 22.04) for me:

```
src/background-filter.cpp:215:31: error: ‘OrtSessionOptionsAppendExecutionProvider_DML’ was not declared in this scope; did you mean ‘OrtSessionOptionsAppendExecutionProvider_CPU’?
  215 |             Ort::ThrowOnError(OrtSessionOptionsAppendExecutionProvider_DML(sessionOptions, 0));
      |                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                               OrtSessionOptionsAppendExecutionProvider_CPU
```
In earlier versions the part with using either `OrtSessionOptionsAppendExecutionProvider_DML` or `OrtSessionOptionsAppendExecutionProvider_CUDA` was used only on WIN32, but now it is also used for linux.

I'm on a AMD GPU and neither have CUDA nor DML, so it throws that error. I've fixed it like this:

```
    try {
#ifdef WITH_CUDA
        if (tf->useGPU == USEGPU_CUDA) {
            Ort::ThrowOnError(OrtSessionOptionsAppendExecutionProvider_CUDA(sessionOptions, 0));
        }
#elif _WIN32
        if (tf->useGPU == USEGPU_DML) {
            Ort::ThrowOnError(OrtSessionOptionsAppendExecutionProvider_DML(sessionOptions, 0));
        }
#endif
        tf->session.reset(new Ort::Session(*tf->env, tf->modelFilepath, sessionOptions));
    } catch (const std::exception& e) {
        blog(LOG_ERROR, "%s", e.what());
        return;
    }
```

That worked for me, but not sure if that is a proper solution. Compilation now works, but OBS segfaults, where I think that is a different issue with the new plugin system in OBS 28.x

sorenisanerd:
This fixes compilation under Linux

Fixes #114 and #125

### DIFF
--- a/src/background-filter.cpp
+++ b/src/background-filter.cpp
@@ -210,7 +210,7 @@ static void createOrtSession(struct background_removal_filter *tf) {
         if (tf->useGPU == USEGPU_CUDA) {
             Ort::ThrowOnError(OrtSessionOptionsAppendExecutionProvider_CUDA(sessionOptions, 0));
         }
-#else
+#elif _WIN32
         if (tf->useGPU == USEGPU_DML) {
             Ort::ThrowOnError(OrtSessionOptionsAppendExecutionProvider_DML(sessionOptions, 0));
         }


tox2ik:
works for me.

royshil:
Implemented on `master` now from a different PR

PR #128, opened 2022-11-30
GPU support on macOS by Vision.framework

umireon:
Apple made Vision.framework and it uses GPU so that the background removal by Vision.framework on macOS is super efficient.

I implemented the code with Vision.framework and thus you can use Apple GPU for the background removal now.

You can install the latest binary from here https://github.com/umireon/obs-backgroundremoval/releases/tag/v0.4.4-umireon

スクリーンショット 2022-11-30 12 27 09


royshil:
@umireon do you want to merge main into this and update it so we can ship?

Issue #67, opened 2021-12-29
Package as a snap or flatpak for a simple user friendly installation on Linux


heidiwenger:
For an easy installation, say, clicking "Install" at Snap Store in Ubuntu, which is the first Linux distro for many new users now coming to Linux. Solus and Manjaro support snaps AND flatpak by default. Fedora defaults flatpak i think, Linux Mint too.

As i read here, many seemed to end up in the world of pain attempting to build from source. Maybe it is time to attempt a creation of a simple pre-package? I mean, why not aim for a simple user friendly way on getting obs-backgroundremoval to go on Linux?

Could this by any chance be possible? What other options there are apart from snaps and flatpaks?

Thank you.

anumerico:
I am having trouble too, getting segmentation fault in ubuntu 20.04.

tam7t:
I put together a flatpak definition that seems to work for v0.4.0: https://gist.github.com/tam7t/dfc8d08ee1042de52921f2cd8a0ce2db

I start with OBS installed through flatpak (`com.obsproject.Studio`, https://feaneron.com/2022/02/14/obs-studio-27-2-on-flathub-get-it-while-its-hot/) then build/install these files as a plugin:

```
$ flatpak-builder build-dir com.obsproject.Studio.Plugin.backgroundremoval.yaml --force-clean
$ flatpak-builder --user --install --force-clean build-dir com.obsproject.Studio.Plugin.backgroundremoval.yaml
```

arpagon:
Nice, thanks @tam7t  [This](https://github.com/royshil/obs-backgroundremoval/issues/67#issuecomment-1140530459) 
 work as charm. 
@royshil Is possible to publish the plugin in flathub?

heidiwenger:
> Nice, thanks @tam7t [This](https://github.com/royshil/obs-backgroundremoval/issues/67#issuecomment-1140530459) work as charm. @royshil Is possible to publish the plugin in flathub?

Oh yes that would be the optimal for the end user (and prolly for everyone else) and in my mind starting this thread.

Three files are required to publish on Flathub:

1. The Flatpak manifest containing the sources and dependencies of the plugin. This file may be either YAML or JSON.

2. AppStream metadata with useful information about your plugin, such as the name, a description, and the license.

3. flathub.json with Flathub build options.

percybolmer:
I just tried this on an Ubuntu and got 

```bash
[...]

PR #147, opened 2023-02-17
Add a flatpak build workflow to this plugin

bergmannf:
I used the files from @tam7t from #67 and added them to a github workflow.

I didn't add any triggers, because right now the workflow will upload the `flatpak` build as an artifact to each run, so I didn't want to clutter your repository with the big artifacts.

But this way it should be reasonably simply to trigger a build every once in a while, so flatpak users can download a plugin they can just install and use.

Or - if nothing else - maybe you can just use this as starting point to create your own better workflow.

Only weird idiosyncrasy I found is that I couldn't use the flatpak bundle feature of the action, as the plugin is a `runtime` not an `app` - that's why I added a step that actually builds the bundle.

### DIFF
--- a/.github/workflows/flatpak.yml
+++ b/.github/workflows/flatpak.yml
@@ -13,8 +13,9 @@ jobs:
   build:
     name: Build flatpak
     runs-on: [ubuntu-latest]
-    env:
-      FLATPAK_BUILD_PATH: flatpak_app/files/share
+    strategy:
+      matrix:
+        arch: [x86_64]
     container:
       image: bilelmoussaoui/flatpak-github-actions:kde-6.4
       options: --privileged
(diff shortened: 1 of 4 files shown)


lcdss:
How do we install this plugin on flatpak? I've tried many different things but nothing new appears on OBS.

bergmannf:
In theory you just download the build artifact from the flatpak action (https://github.com/royshil/obs-backgroundremoval/actions/runs/4268092426) - unzip it and install it using `flatpak install --user <the_unzipped_file>`.

Once that is done you should be able to use the plugin normally.

ferretiz:
Hi @bergmannf 
I just tested your instructions to install the plugin and it is working like a charm!
I tested on:
Ubuntu 22.04.2 LTS 64 bit
Flatpak 1.12.7
OBS Studio com.obsproject.Studio  29.0.2

Thanks a lot!!

PR #157, opened 2023-03-08
Add CoreML support for Mac

royshil:
To boost inference time and save on CPU cycles on Mac machines

### DIFF
--- a/cmake/BuildMyOnnxruntime.cmake
+++ b/cmake/BuildMyOnnxruntime.cmake
@@ -43,6 +43,7 @@ elseif(OS_MACOS)
       ${CMAKE_OSX_DEPLOYMENT_TARGET}
       --osx_arch
       ${CMAKE_OSX_ARCHITECTURES}
+      --use_coreml
       --cmake_extra_defines
       CMAKE_C_COMPILER_LAUNCHER=ccache
       --cmake_extra_defines


umireon:
@royshil Can you try #160 which is a PR against this branch?

umireon:
It would be great if the version number in CMakeLists.txt and buildspec.json is changed to the correct number before the new release

royshil:
> It would be great if the version number in CMakeLists.txt and buildspec.json is changed to the correct number before the new release

so v0.5.1 or 0.5.2?

umireon:
I think v0.5.2

umireon:
Builds seem to be fixed 🎉

royshil:
@umireon yep but it's not working 🤣  .. i'm fixing - it should only be a minute longer.

src/background-filter.cpp
```diff
@@ -109,9 +111,13 @@ static obs_properties_t *filter_properties(void *data)
   obs_property_list_add_string(p_use_gpu, obs_module_text("CPU"), USEGPU_CPU);
 #ifdef WITH_CUDA
   obs_property_list_add_string(p_use_gpu, obs_module_text("GPU - CUDA"), USEGPU_CUDA);
-#elif _WIN32
+#endif
+#if _WIN32
   obs_property_list_add_string(p_use_gpu, obs_module_text("GPU - DirectML"), USEGPU_DML);
 #endif
+#if APPLE
```
```suggestion
#ifdef __APPLE__
```

umireon:
With CoreML:
スクリーンショット 2023-03-08 14 43 22

With CPU:
スクリーンショット 2023-03-08 14 43 58

Using CoreML results in the FPS deterioration 🤔

royshil:
> With CoreML: スクリーンショット 2023-03-08 14 43 22
> 
> With CPU: スクリーンショット 2023-03-08 14 43 58
> 
> Using CoreML results in the FPS deterioration 🤔

yes i noticed that too.. and it's not gaining anything in CPU reduction or increasing GPU...

umireon:
I suppose almost all the operation of OBS is done with GPU.
GPU is not free but CPU is free so it is better to use CPU for additional computing on Mac.

umireon:
CoreML uses CPU when the operator is not supported by CoreML.
Once I investigated this and found that half of the operators In MediaPipe were executed in CPU.

royshil:
@umireon coreML also does nothing on Intel Macs. Only supported on M1/M2. Which is fine.
we can land this as is, and let the people choose...

umireon (approved):
LGTM!

royshil:
@umireon can you plz approve? thanks!

umireon (approved):
Oh my approval discarded, LGTM!

PR #158, opened 2023-03-08
Enable -Werror and /WX

umireon:
Enables -Werror in linux and mac and /WX in Windows.

The increment in assignment may conclude in undefined behavior.
std::pow is double so they turned into integer operations.


royshil (approved):
LGTM! well done

PR #166, opened 2023-03-09
Add PPHumanSeg new model and background blur

royshil:
Adding the model from https://github.com/PaddlePaddle/PaddleSeg/tree/release/2.7/contrib/PP-HumanSeg
As well as a function to blur the background instead of removing with a color


src/models/Model.h
```diff
@@ -38,6 +38,26 @@ static void hwc_to_chw(cv::InputArray src, cv::OutputArray dst)
...
+
+  // for each channel, extract a HxW image and store it in the vector
+  std::vector<cv::Mat> channelsMat;
+  for (int i = 0; i < channels; i++) {
+    channelsMat.push_back(src.row(i).reshape(1, height));
+  }
+
+  // merge the channels together
+  cv::merge(channelsMat, dst);
+}
```
umireon:
```suggestion
// static void chw_to_hwc(cv::InputArray src_, cv::OutputArray dst)
// {
//   cv::Mat src = src_.getMat();
//
//   // convert a CHW (channels, height, width) image to HWC (height, width, channels)
//   // determine the number of channels
//   int channels = src.size().height;
//   // determine the height and width of the image
//   int height = src.size().width;
//
//   // for each channel, extract a HxW image and store it in the vector
//   std::vector<cv::Mat> channelsMat;
//   for (int i = 0; i < channels; i++) {
//     channelsMat.push_back(src.row(i).reshape(1, height));
//   }
//
//   // merge the channels together
//   cv::merge(channelsMat, dst);
// }
```

src/background-filter.cpp
```diff
@@ -261,7 +273,9 @@ static void filter_update(void *data, obs_data_t *settings)
   tf->smoothContour = (float)obs_data_get_double(settings, "smooth_contour");
   tf->feather = (float)obs_data_get_double(settings, "feather");
   tf->maskEveryXFrames = (int)obs_data_get_int(settings, "mask_every_x_frames");
+  tf->maskEveryXFrames = (int)obs_data_get_int(settings, "mask_every_x_frames");
   tf->maskEveryXFramesCount = (int)(0);
+  tf->blurBackground = (float)obs_data_get_int(settings, "blur_background");
```
```suggestion
  tf->blurBackground = obs_data_get_int(settings, "blur_background");
```

umireon (changes requested):
Fix errors

umireon (approved):
LGTM!

PR #169, opened 2023-03-10
Fix Build error

RikvdHeijden:
Building on Pop_os resulted in an At least one entry of URL is a path (invalid in a list) warning. This seems to be caused by a forgotten newline in this function. I'm no expert at this, so please correct me if I'm wrong, but this change got my build working again.

### DIFF
--- a/cmake/BuildMyOpenCV.cmake
+++ b/cmake/BuildMyOpenCV.cmake
@@ -33,8 +33,8 @@ endif()
 
 ExternalProject_Add(
   OpenCV_Build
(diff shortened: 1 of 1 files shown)


umireon:
I think old CMake does not recognize DOWNLOAD_EXTRACT_TIMESTAMP.
It would be great if we bump cmake_minimum_version

royshil (approved):
just a formatting change
LGTM

umireon (approved):
LGTM

PR #174, opened 2023-03-13
Bump 0.5.14

umireon:


royshil (approved):
may want to merge this into locale PR... anyway this is good.

PR #180, opened 2023-03-14
Add flatpak build to CI

royshil:


umireon:
I think flatpak should be buiit on the separate job. Messing Linux deb job is not good.

umireon:
I can work on this. I have the environment to develop the flatpak package.

royshil:
@umireon that's great
i dn't have a linux dev env.. only windows and mac
i think the latest change to the cmake script with shared libraries should enable flatpak pretty easy

umireon:
Main cannot be built on flatpak due to #182

umireon:
My work done 
https://github.com/royshil/obs-backgroundremoval/pull/198

PR #185, opened 2023-03-16
Add USE_SHARED_OPENCV and USE_SHARED_ONNXRUNTIME options

umireon:
Flatpak doesn't like ExternalProject.
This change will help us to use pre-built binaries and separately built libraries for Flatpak.
Pkgs, exe, deb will remain to use statically-linked libraries.


CMakeLists.txt
```diff
@@ -45,8 +46,43 @@ configure_file(src/plugin-macros.h.in ${CMAKE_SOURCE_DIR}/src/plugin-macros.gene
...
+    add_library(Onnxruntime SHARED IMPORTED)
+    set_target_properties(Onnxruntime PROPERTIES IMPORTED_LOCATION ${Onnxruntime_LIBRARIES})
+    set(Onnxruntime_INCLUDE_PATH
+        ${Onnxruntime_INCLUDE_DIR} ${Onnxruntime_INCLUDE_DIR}/onnxruntime
+        ${Onnxruntime_INCLUDE_DIR}/onnxruntime/core/session
+        ${Onnxruntime_INCLUDE_DIR}/onnxruntime/core/providers/cpu)
+    set_target_properties(Onnxruntime PROPERTIES INTERFACE_INCLUDE_DIRECTORIES
+                                                 "${Onnxruntime_INCLUDE_PATH}")
+  else()
+    message(FATAL_ERROR "Only on Linux the shared ONNX Runtime supported!")
```
royshil: nit: "Shared ONNX Runtime is only supported on Linux!"

CMakeLists.txt
```diff
@@ -45,8 +46,43 @@ configure_file(src/plugin-macros.h.in ${CMAKE_SOURCE_DIR}/src/plugin-macros.gene
...
+    OFF
+    CACHE STRING "Use system OpenCV")
+if(USE_SHARED_OPENCV)
+  if(OS_LINUX)
+    find_package(OpenCV REQUIRED)
+    add_library(OpenCV SHARED IMPORTED)
+    set_target_properties(OpenCV PROPERTIES IMPORTED_LOCATION ${OpenCV_LIBRARIES})
+    set_target_properties(OpenCV PROPERTIES INTERFACE_INCLUDE_DIRECTORIES "${OpenCV_INCLUDE_DIRS}")
+  else()
+    message(FATAL_ERROR "Only on Linux the shared OpenCV supported!")
```
royshil: same
"Shared OpenCV is only supported on Linux!"

CMakeLists.txt
```diff
@@ -45,8 +46,43 @@ configure_file(src/plugin-macros.h.in ${CMAKE_SOURCE_DIR}/src/plugin-macros.gene
 
 target_sources(${CMAKE_PROJECT_NAME} PRIVATE src/plugin-macros.generated.h)
 
-include(cmake/BuildMyOnnxruntime.cmake)
-include(cmake/BuildMyOpenCV.cmake)
+set(USE_SHARED_ONNXRUNTIME
+    OFF
```
royshil, line 50: well instead of `OFF` here you can simply `${OS_LINUX}` and that will turn it on for Linux and off for the others

CMakeLists.txt
```diff
@@ -45,8 +46,43 @@ configure_file(src/plugin-macros.h.in ${CMAKE_SOURCE_DIR}/src/plugin-macros.gene
...
+else()
+  include(cmake/BuildMyOnnxruntime.cmake)
+endif()
+
+set(USE_SHARED_OPENCV
+    OFF
+    CACHE STRING "Use system OpenCV")
+if(USE_SHARED_OPENCV)
+  if(OS_LINUX)
+    find_package(OpenCV REQUIRED)
```
royshil: we should limit the components here to only what we need which is `core imgproc`
so you can add `COMPONENTS core imgproc`

CMakeLists.txt
```diff
@@ -45,8 +46,43 @@ configure_file(src/plugin-macros.h.in ${CMAKE_SOURCE_DIR}/src/plugin-macros.gene
...
 target_sources(${CMAKE_PROJECT_NAME} PRIVATE src/plugin-macros.generated.h)
 
-include(cmake/BuildMyOnnxruntime.cmake)
-include(cmake/BuildMyOpenCV.cmake)
+set(USE_SHARED_ONNXRUNTIME
+    OFF
+    CACHE STRING "Use system ONNX Runtime")
+if(USE_SHARED_ONNXRUNTIME)
+  if(OS_LINUX)
+    find_package(Onnxruntime REQUIRED)
```
royshil: i wonder if we should constrain for a version e.g. 1.13 or 1.14 here

royshil (changes requested):
some changes to take care of

CMakeLists.txt
umireon: I meant that using shared libraries is only for flatpak distribution and I want to keep the deb distribution don't require additional libraries.
I maybe want to rename the flags to USE_SYSTEM_FOO.

royshil (approved):
looks good now
merge at will

PR #186, opened 2023-03-17
Abort if tf->model is null in createOrtSession

### LINKED ISSUE #184: Windows 11 pro Antivirus removes this package
image

umireon:
If a user have migrated from an older version of this plugin, old model files may have remained and tf->modelSelection is set not to be prefixed with models. This change aborts the createOrtSession if tf->model is null and it prevents the segmentation fault of executing tf->model->populateInputOutputNames() when tf->model is null.

Fixes #184.


royshil (approved):
nice ! thanks for the fix

Issue #101, opened 2022-07-01
Chroma key with feather filter creates some color spill for me


claell:
I want to use this plugin with a virtual background. For that, it is suggested to use a chroma key filter.

However, together with the "Feather blend silhouette" option of this, the chroma key filter creates a color spill on the edges, which doesn't look nice:

![grafik](https://user-images.githubusercontent.com/6171135/176916159-52de5c80-3dc1-4605-b2d0-3e7344b55d40.png)

If I don't use the "Feather blend silhouette", the edges don't look good. I didn't find an option to later add the feather blend (for example, after or in chroma key filter).

Ideally, one could simply make the background transparent in this filter without the need of additional chroma key filter in the chain. Unfortunately, currently, the color picker doesn't give me that option.

Do you know a good solution or possibly can you add some?

For now, I just reduce the strength of "Feather blend silhouette" as a workaround.

royshil:
last time i checked transparency wasn't possible.. had to pair it with a chroma key...
we'll look into the color bleed for sure

royshil:
color bleed should be fixed in latest version coming out

PR #190, opened 2023-03-19
Background removal without Chrome Key filter

umireon:
TODO: implement feather, configuration

Possible solution for #101

### DIFF
--- a/src/background-filter.cpp
+++ b/src/background-filter.cpp
@@ -78,7 +78,6 @@ struct background_removal_filter {
   int maskEveryXFrames = 1;
   int maskEveryXFramesCount = 0;
   int64_t blurBackground = 0;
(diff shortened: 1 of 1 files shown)


umireon:
I suggest that this change would go with v0.6.0.

umireon:
@royshil Could you help me to implement the feather matting with the alpha channel?

royshil:
@umireon sure i can - but this branch is on your repo, so i'd need permissions there to be able to push to it

another option is to land this change, and i'll do the feather on this repo's branch

umireon:
I've moved the branch on royshil/obs-backgroundremoval and opened https://github.com/royshil/obs-backgroundremoval/pull/192 so that you can write to my PR.

Issue #195, opened 2023-03-20
obs: symbol lookup error: /usr//lib/x86_64-linux-gnu/obs-plugins/obs-backgroundremoval.so: undefined symbol: cpuinfo_initialize


piphoser:
OBS 29.0.2 from OBS ppa:obsproject/obs-studio repository crashes when adding the backgroundremoval filter with the above error message:  obs: symbol lookup error: /usr//lib/x86_64-linux-gnu/obs-plugins/obs-backgroundremoval.so: undefined symbol: cpuinfo_initialize

Is there a dependency I'm missing?

royshil:
yes apparently something is wrong
we're looking into it
thanks for the issue

this is a duplicate of #194

royshil:
@piphoser  can you try https://github.com/royshil/obs-backgroundremoval/suites/11693307176/artifacts/608110162 ?

royshil:
this seems to be resolved

PR #196, opened 2023-03-21
Disable cpuinfo in Linux to prevent crash

### LINKED ISSUE #194: UBUNTU 22.04, OBS 29 installing latest Plug-In obs-backgroundremoval-0.5.14-linux-x86_64.deb
**Describe the bug**
obs: symbol lookup error: /usr//lib/x86_64-linux-gnu/obs-plugins/obs-backgroundremoval.so: undefined symbol: c
puinfo_initialize

**To Reproduce**
Start OBS from a Terminal Window. OBS fails to start with above error.
If started normally, OBS just terminates w/o any notification.
No issue with 0.5.13 Version

**Expected behavior**
I expect OBS to start

 - OS: UBUNTU 22.04 (current on Updates)
 - Plugin Version 0.5.14
- OBS Version 29

**Additional context**

royshil:
trying to resolve #194 #195

### DIFF
--- a/cmake/BuildMyOnnxruntime.cmake
+++ b/cmake/BuildMyOnnxruntime.cmake
@@ -68,7 +68,8 @@ else()
   if(Onnxruntime_CCACHE_EXE)
     list(APPEND Onnxruntime_PLATFORM_OPTIONS --cmake_extra_defines
          CMAKE_C_COMPILER_LAUNCHER=${Onnxruntime_CCACHE_EXE} --cmake_extra_defines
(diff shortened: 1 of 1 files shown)


umireon:
I rather think we should link the missing library than remove cpuinfo.

royshil:
@umireon yes i thought so too, but we'd need to statically link it which is a pain
they have a fallback in ONNX RT for not using cpuinfo, i don't think it's a big deal if it's missing

royshil:
@umireon do you thinkw e should merge or do you want to find a solution for cpuinfo?
also we must verify this actualy fixes the problem

umireon:
Let me check if the linkage error can be fixed easily

PR #199, opened 2023-03-21
Correct LICENSE to GPL-2.0

umireon:
I know you don't like GPL but since we are linking to the GPL-2.0 product named OBS, we have to obey the GPL-2.0 license.
I don't assume that publishing our product under MIT License is permitted under GPL-2.0 terms and I suppose all of the OBS plugins on the earth must be GPL-2.0.

> GNU GPLv2
>
> The GNU GPL is the most widely used free software license and has a strong copyleft requirement. When distributing derived works, the source code of the work must be made available under the same license. There are multiple variants of the GNU GPL, each with different requirements.

https://choosealicense.com/licenses/gpl-2.0/

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,339 @@
-MIT License
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 2, June 1991
 
-Copyright (c) 2021 Roy Shilkrot
+ Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
+ 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
 
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-copies of the Software, and to permit persons to whom the Software is
-furnished to do so, subject to the following conditions:
+                            Preamble
 
-The above copyright notice and this permission notice shall be included in all
-copies or substantial portions of the Software.
+  The licenses for most software are designed to take away your
+freedom to share and change it.  By contrast, the GNU General Public
+License is intended to guarantee your freedom to share and change free
+software--to make sure the software is free for all its users.  This
+General Public License applies to most of the Free Software
+Foundation's software and to any other program whose authors commit to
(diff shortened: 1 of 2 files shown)


royshil:
i don't know if this is true
we are not modifying OBS code, this is an external plugin

can you find examples for GPL2 restriction on plugins?
i'm trying to read around to educate myself, but i can't find any instructions on plugins, or even discussion in the forum

umireon:
> we are not modifying OBS code, this is an external plugin

Unfortunately, no, GPL thinks our work is a part of OBS.

https://www.gnu.org/licenses/gpl-faq.html.en#GPLStaticVsDynamic

> Linking a GPL covered work statically or dynamically with other modules is making a combined work based on the GPL covered work. Thus, the terms and conditions of the GNU General Public License cover the whole combination.

https://www.gnu.org/licenses/gpl-faq.html.en#TheGPLSaysModifiedVersions

> Section 2 says that modified versions you distribute must be licensed to all third parties under the GPL.

> can you find examples for GPL2 restriction on plugins?

The restriction of GPL-2.0 is that the plugins can not virtually be sold because the source code must be public under GPL-2.0.
https://obsproject.com/forum/threads/obs-license-infringement.141405/

royshil:
> > we are not modifying OBS code, this is an external plugin
> 
> Unfortunately, no, GPL thinks our work is a part of OBS.
> 
> https://www.gnu.org/licenses/gpl-faq.html.en#GPLStaticVsDynamic
> 
> > Linking a GPL covered work statically or dynamically with other modules is making a combined work based on the GPL covered work. Thus, the terms and conditions of the GNU General Public License cover the whole combination.
> 
> https://www.gnu.org/licenses/gpl-faq.html.en#TheGPLSaysModifiedVersions
> 
> > Section 2 says that modified versions you distribute must be licensed to all third parties under the GPL.
> 
> > can you find examples for GPL2 restriction on plugins?
> 
> The restriction of GPL-2.0 is that the plugins can not virtually be sold because the source code must be public under GPL-2.0. https://obsproject.com/forum/threads/obs-license-infringement.141405/

ok you make a valid point.

i'm not dying on this hill, so let's change the license to GPLv2.

royshil (approved):

umireon:
@royshil I appreciate your kind understanding because this change is hard to accept 🙇 
Once I search for a way to avoid releasing my plugin under GPL-2.0 but I found there was no way and I was very disappointed....

umireon:
@royshil The author of this plugin is you and thus you should make this change.
There might be a legal problem if a non-author made this commit to change its license.
Could you please create another PR with the appropriate changes and then I will approve it?

PR #202, opened 2023-03-21
Remove BG color, transparent by default

royshil:
Removing the ability to set the background color, it will always be transparent.
This will undo the need for an additional chroma key filter.

closing #193

### DIFF
--- a/src/background-filter.cpp
+++ b/src/background-filter.cpp
@@ -486,7 +486,7 @@ static struct obs_source_frame *filter_render(void *data, struct obs_source_fram
       // bgra[0,1,2] -> bgra[0,1,2],
       // bgra[3] -> alpha[0]
       int from_to[] = {0, 0, 1, 1, 2, 2, 3, 3};
-      mixChannels(&bgra, 1, out, 2, from_to, 4);
+      mixChannels(&imageBGRA, 1, out, 2, from_to, 4);
     } else {
       // If we're not feathering/alpha blending, we can
       // apply the mask as-is back onto the main image.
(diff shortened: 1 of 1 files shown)


src/background-filter.cpp
```diff
@@ -481,40 +472,21 @@ static struct obs_source_frame *filter_render(void *data, struct obs_source_fram
...
-          }
-        }
-      }
+      cv::Mat alpha;
+      cv::Mat((cv::Scalar(1.0) - maskFloat) * 255.0).convertTo(alpha, CV_8UC1);
+      cv::Mat out[] = {imageBGRA, alpha};
+      // bgra[0,1,2] -> bgra[0,1,2],
+      // bgra[3] -> alpha[0]
+      int from_to[] = {0, 0, 1, 1, 2, 2, 3, 3};
+      mixChannels(&imageBGRA, 1, out, 2, from_to, 4);
```
```suggestion
      cv::Mat in[] = {imageBGRA, alpha};
      // bgra[0,1,2] -> bgra[0,1,2],
      // alpha[0] -> bgra[3]
      int from_to[] = {0, 0, 1, 1, 2, 2, 4, 3};
      mixChannels(in, 2, &imageBGRA, 1, from_to, 4);
```

umireon (changes requested):

src/background-filter.cpp
umireon: I think src and dst are inverted.

umireon:
```suggestion
      // alpha[0] -> bgra[3]
      int from_to[] = {0, 3};
      mixChannels(&alpha, 1, &imageBGRA, 1, from_to, 1);
```

Maybe better

umireon (approved):
LGTM!

PR #203, opened 2023-03-21
Fix CMakeLists.txt for Flatpak builds

umireon:
Flatpak has CMake 3.25.1.
cmake_minimum_required must be below 3.25.1 for Flatpak.
Building ONNX Runtime requires 3.25.2 but we use pre-built ONNX Runtime for Flatpak so that CMake 3.25.1 is valid only for Flatpak build.
The deb build still requires 3.25.2 and this is great to note on README.md or CONTRIBUTING.md.


royshil (approved):
looking good!

PR #204, opened 2023-03-21
Feathering should be done within the contour

umireon:
In the current implementation, feathering the portrait will include the edge of the background and I don't think this is what people expect from the feathering feature.
スクリーンショット 2023-03-22 4 08 30

Adding erosion process before boxFilter will be what people expect.
スクリーンショット 2023-03-22 4 10 12

### DIFF
--- a/src/background-filter.cpp
+++ b/src/background-filter.cpp
@@ -487,6 +487,7 @@ static struct obs_source_frame *filter_render(void *data, struct obs_source_fram
       // Convert Mat to float and Normalize the alpha mask to [0,1].
       tf->backgroundMask.convertTo(maskFloat, CV_32FC1, 1.0 / 255.0);
       // Feather (blur) the normalized mask
+      cv::dilate(maskFloat, maskFloat, cv::Mat(), cv::Point(-1, -1), k_size / 3);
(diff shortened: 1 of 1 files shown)


src/background-filter.cpp
```diff
@@ -487,6 +487,7 @@ static struct obs_source_frame *filter_render(void *data, struct obs_source_fram
       // Convert Mat to float and Normalize the alpha mask to [0,1].
       tf->backgroundMask.convertTo(maskFloat, CV_32FC1, 1.0 / 255.0);
       // Feather (blur) the normalized mask
+      cv::dilate(maskFloat, maskFloat, cv::Mat(), cv::Point(-1, -1), k_size / 3);
```
royshil, line 490: i wonder if we should make the dilation factor a parameter? or we just use this "magic number" of `k/3` ..

either way i think this is a good idea!

royshil (approved):
lgtm!

src/background-filter.cpp
umireon: I want to keep it automatic.
People will want to remove the edge of the background but not want to bother with a new parameter.

PR #205, opened 2023-03-22
Remove opencv dependencies from Aptfile

umireon:
OpenCV libraries will be built and not be required on the system.


royshil (approved):
go!

Issue #146, opened 2023-02-01
Cannot install ob Mac


jomach:
I cannot install this plugin on a mac with obs 29 x64. I copied as instructed and it did not work

thomasfredericks:
I concur. OBS macOS Intel 29.0.2 does not find the latest obs-backgroundremoval even if the installation instructions are followed.

cryptotoolslive:
I am also haveing the same issue with 20.0.2 x64 on MacOS 13.2.

UsghuiYz:
I'm having the same issue. It simply does not show up in the filters list.

PhilippeBourque:
Same here. Thanks

dkp:
@royshil 
OSX Monterey 12.6.3
OBS 29.0.2

I'm assuming the instructions are:
Mac OSX: Unzip obs-backgroundremoval-macosx.zip to /Applications/OBS.app/Contents/. e.g. $ unzip -o obs-backgroundremoval-macosx.zip -d /Applications/OBS.app/Contents/

However, the main github page provides instructions for building the plugin.  I assume this is not necessary??

I do not see the filter in the filters list.

VForsmann:
Same here for me.
Running on ARM M1 Max.
Also tried OBS v28.0.0 - no filter coming up after extracting the .zip as told in instructions.

royshil:
I'm rebuilding the package and plugin as a whole. Already made progressNew version coming out very soonIt will have a .pkg installer instead of the .zipOn Feb 26, 2023, at 12:20 PM, Vitus Forsmann ***@***.***> wrote:﻿
Same here for me.
Running on ARM M1 Max.
Also tried OBS v28.0.0 - no filter coming up after extracting the .zip as told in instructions.

pas4gok:
Similar issue unable to find the plugin after following instructions to unarchive to the folder as mentioned. Running on 2018 MB Pro OBS 29.0.2

TheFutonEng:
This is also not working on M1 Mac running Monterey 12.4.

dayojohn19:
not yet resolved?

OBS1227:
I just downloaded the latest version 0.5.0 Background Remover and installed it via the  installer, but it doesn't show up in OBS 29.2 at all. I did an OBS restart and the same. I had successfully installed it manually in older versions of OBS, but the menus have changed from the older versions of OBS, so I haven't been able to install manually or via the installer for the later versions.

[...]

PR #208, opened 2023-03-23
Use pkgbuild and productbuild for .pkg

umireon:
The pkg file created by the Packages will require users to do the following steps.

https://github.com/royshil/obs-backgroundremoval/issues/146#issuecomment-1456343517

The pkgbuild and productbuild commands are the newer way to create pkgs and AFAIK pkgs created by these commands won't require users to take the additional steps to install files under the home directory.

### DIFF
--- a/.github/scripts/.package.zsh
+++ b/.github/scripts/.package.zsh
@@ -155,6 +155,7 @@ Usage: %B${functrace[1]%:*}%b <option> [<options>]
       --distribution "${project_root}/cmake/bundle/macos/Distribution.xml" \
       --package-path "${project_root}/release" \
       "${project_root}/release/${product_name}.pkg"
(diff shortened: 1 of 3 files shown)


royshil:
@umireon this LGTM but i can't make it build the `.pkg` so i can verify it works...

umireon:
@royshil You can check if the change works by executing `zsh ./.github/scripts/package-macos.zsh` locally and you will get the pkg under the release directory.
CI won't create pkgs until we make a new release even if the Seeking Testsers label is added to PR.

umireon:
The previous pkg does not work as expected when codesigned and I added the postinstall script to ensure the plugin is installed in ~/Library/Application Support/obs-studio/plugins even if the installation location is / by the Apple's bug.

royshil (approved):
this is looking good
let's merge it and release a new version with the signing

Issue #206, opened 2023-03-22
Remove or compress the MODNET model


royshil:
**Is your feature request related to a problem? Please describe.**
the MODNET model is the biggest one we carry at 20Mb, it's not the default and it's not that good.
the package is 50-70Mb today, we can bring it way down.

**Describe the solution you'd like**
either we remove it or we compress it with some e.g. quantization

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #209, opened 2023-03-24
Remove MODNET model

royshil:
The MODNET model is heavy in size (20Mb) and it's accuracy is low compared to the others.

we have pretty good models already, so we can cut back

closing #206


cmake/BuildMyOnnxruntime.cmake
umireon, line 1: I suppose your changes in BuildMyOnnxruntime.cmake would break the Linux build again.

umireon (changes requested):

cmake/BuildMyOnnxruntime.cmake
umireon: libonnxruntime_session.a must come first in CMake.

umireon: I've confirmed that these changes in BuildMyOnnxRuntime.cmake broke the Linux build.

スクリーンショット 2023-03-24 16 53 45

royshil:
@umireon i was mistaken. i accidentally checked in the Cmake changes - where i shouldnt
but i've reverted hthem now
can you check it out please?

umireon (approved):
LGTM!

PR #211, opened 2023-03-26
Adopt stale bot

umireon:
Handling stale issues automatically.
We have to create no-issue-activity label for the bot to check if 14 days passed after the issue was marked as stalled.
If we adopt the stall bot, the criteria of our closing issue would be clear and the closed issue participants would be convinced.

https://github.com/marketplace/actions/close-stale-issues


royshil (approved):
perfect! well done

PR #213, opened 2023-03-26
Data in obs_source_frame must belong to the same bzalloc memory.

### LINKED ISSUE #212: 0.5.15 cannot work with the color formats which has more than one plane
**Describe the bug**
When applying the `Background Removal` filter from this plugin to any source, the source will disappear until the filter is removed or hidden. This will cause a green video until the source is hidden and shown.

Additionally, the `Background Color` property seems to be missing from the latest release.

**To Reproduce**
1. Add a `Background Removal` filter to any source
2. Make sure it's enabled
3. Source output will disappear

**Expected behavior**

**Screenshots**

https://user-images.githubusercontent.com/86388887/227756226-000aed23-d9d2-4b64-bd11-5783350ce4b5.mp4

- OS: Windows 10 (22H2)
- Browser: N/A
- Plugin Version: v0.5.15
- OBS Version: v29.0.2
- GPU: NVIDIA GeForce GTX 1650 (Max-Q)

**Additional context**
I have reproduced this bug with both a `Media Source` and `Video Capture Device`, with any `Segmentation Model`, `Inference device`, `Blur Background Factor` or `Calculate Mask every X frame` setting.

umireon:
The data field in obs_source_frame must belong to the same bzalloc otherwise OBS references invalid memory area.

### DIFF
--- a/src/background-filter.cpp
+++ b/src/background-filter.cpp
@@ -77,9 +77,6 @@ struct background_removal_filter {
   int maskEveryXFramesCount = 0;
   int64_t blurBackground = 0;
 
-  uint8_t *previousFrameData0;
-  int previousFrameLinesize0;
-
 #if _WIN32
   const wchar_t *modelFilepath = nullptr;
 #else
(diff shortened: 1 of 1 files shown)


src/background-filter.cpp
```diff
@@ -496,10 +496,17 @@ static struct obs_source_frame *filter_render(void *data, struct obs_source_fram
...
 
-  bfree(frame->data[0]);
-  obs_source_frame_init(frame, VIDEO_FORMAT_BGRA, imageBGRA.cols, imageBGRA.rows);
-  std::memcpy(frame->data[0], imageBGRA.data,
-              imageBGRA.cols * imageBGRA.elemSize() * imageBGRA.rows);
+  uint8_t *data0 = frame->data[0];
+  uint8_t *data1 = frame->data[1];
+  uint8_t *data2 = frame->data[2];
+  uint8_t *data3 = frame->data[3];
+  frame->data[0] = static_cast<uint8_t *>(brealloc(data0, imageBGRA.cols * imageBGRA.rows * 16));
```
royshil, line 503: why do we have to alloc `* 16` here? shouldn't it be just `*4` for RGBA?

also, `imageRGBA` is packed `RGBARGBARGBA`
while it seems `frame->data` is planar packed e.g. `RRRRRR..... GGGGGG...... BBBBBBB..... AAAAAAA`
so we can't memcpy?

also, where do we free this memory?

royshil (changes requested):
i'm not sure this is the solution we need..

src/background-filter.cpp
umireon: > why do we have to alloc * 16 here? shouldn't it be just *4 for RGBA?

I checked the OBS source code and I found that the maximum size of a data plane is BGRA's and its size is `width * height * 4`.
The current maximum number of planes is 4 so we have to keep `frame->data[3]` allocated.
Thus I want to allocate width * height * 4 to 4 data planes as a proof of concept.

https://github.com/obsproject/obs-studio/blob/90f46a9d4dda4842325e513f5b3bcdc395a37803/libobs/media-io/video-frame.c#L23

Indeed, we can allocate memory more efficiently but it would need to create a different logic on the input frame->format.
I just want to keep the implementation simple for the first time.

> also, imageRGBA is packed RGBARGBARGBA

No, BGRABGRABGRA...

> while it seems frame->data is planar packed e.g. RRRRRR..... GGGGGG...... BBBBBBB..... AAAAAAA
> so we can't memcpy?

I instructed OBS to read data[0] as BGRABGRABGRA....

```
frame->format = VIDEO_FORMAT_BGRA;
```

ImageBGRA is packed as BGRA and thus imageBGRA.data and data[0] have the same structure as BGRA so they can memcpyed.

> also, where do we free this memory?

OBS will free frame->data[0] when OBS exits.
OBS assumes that frame->data[1+] must be freed by freeing frame->data[0].
Note that I have decided to call brealloc and no new pointers are allocated.

royshil: if we're using only `frame->data[0]` why are we setting the `[1],[2],..` ?
those pointers would be "grabage", they are pointing to a place in the data that doesn't mean much
since all the data is already packed in `data[0]`

also `VIDEO_FORMAT_BGRA` - does it imply `BGRABGRABGRA` packing? or is it planar like `BBB GGG RRR AAA`?

src/background-filter.cpp
```diff
@@ -496,10 +496,17 @@ static struct obs_source_frame *filter_render(void *data, struct obs_source_fram
...
+  frame->data[1] = frame->data[0] + (imageBGRA.cols * imageBGRA.rows * 4);
+  frame->data[2] = frame->data[1] + (data2 - data1);
+  frame->data[3] = frame->data[1] + (data3 - data1);
```
royshil, line 506: if we're not using `data[1], data[2]` etc. - why not simply `frame->data[1] = data1` and so on?

royshil (changes requested):

umireon:
Since we make output with BGRA, the size of data[0] has to be `width * height * 4` and linesize[0] has to be `width * 4`.

OBS reuses `frame` between frames.

I will illustrate the case of NV12.
The size of data[0] must be more than `width` and linesize[0] must be more than `width` with NV12.
The redundant area is simply ignored so it is okay to have more size and more linesize than required.
We want to extend the size of data[0] and the address of data[0] may change after data[0] is reallocated.
When OBS starts the frame, OBS reuses data[0] and data[1] on the `frame` that we returned.
Therefore, the pointer of data[1] must be valid when we return the `frame` on our plugin.
To keep data[1] valid, we have to set a pointer based on the new data[0] to data[1].

I admit this is hacky but at least it works.
I think it would be great if we can output with gs_ functions as the Chroma Key filter does.

umireon:
> also VIDEO_FORMAT_BGRA - does it imply BGRABGRABGRA packing? or is it planar like BBB GGG RRR AAA?

VIDEO_FORMAT_BGRA implies BGRABGRABGRA packing!

Issue #215, opened 2023-03-26
Feather Blend Silhouette cannot be used with Background Blur


Sparronator9999:
**Is your feature request related to a problem? Please describe.**
The latest version (0.15.5) of OBS Background Removal cannot currently apply a Feather Blend to a source if the Blur Background setting is not set to 0. Trying to do this will cause the Feather Blend to override the Background Blur and will cause the background to be removed instead.

Being able to feather blend the blurred background would, in my opinion, improve the quality of the background blur as it reduces the harsh transition from non-blurred to blurred background.

**Describe the solution you'd like**
Make the `Feather Blend Silhouette` and `Background Blur Factor` settings compatible (i.e. apply the feather blend to the source with blurred background)

**Describe alternatives you've considered**
A workaround would be to use a blurred screenshot of the room background (with no-one in frame) and add this as an Image source, then apply the `Background Removal` effect with feather blending to the `Video Capture Device`.

**Additional context**
Screenshots of the issue:

Feather blend on: | Feather blend off:
-------------------|-------------------
![Capture3](https://user-images.githubusercontent.com/86388887/227766630-c9de9f7d-97b4-4442-9098-cbd279591a7a.PNG) | ![Capture2](https://user-images.githubusercontent.com/86388887/227766632-d951512d-14f6-44db-b4a6-cb613e24a4b4.PNG)

(Probably not the best example, since the background is solid. Oh well)

PR #217, opened 2023-03-27
Fix feather + background blur bug

royshil:
Bug reported (verified) from #215

### DIFF
--- a/src/background-filter.cpp
+++ b/src/background-filter.cpp
@@ -435,7 +435,7 @@ void blend_images_with_mask(cv::Mat &dst, const cv::Mat &src, const cv::Mat &mas
 {
   for (int i = 0; i < dst.rows; i++) {
     const cv::Vec4b *srcRow = src.ptr<cv::Vec4b>(i);
-    const cv::Vec4b *maskRow = mask.ptr<uchar>(i);
+    const uchar *maskRow = mask.ptr<uchar>(i);
     cv::Vec4b *dstRow = dst.ptr<cv::Vec4b>(i);
 
     for (int j = 0; j < dst.cols; j++) {
--- a/src/background-filter.cpp
+++ b/src/background-filter.cpp
@@ -443,10 +443,10 @@ void blend_images_with_mask(cv::Mat &dst, const cv::Mat &src, const cv::Mat &mas
       const float maskPixel = (float)(maskRow[j]) / 255.0f;
       cv::Vec4b &dstPixel = dstRow[j];
 
-      if (maskPixel == 0) {
+      if (maskPixel == 1.0) {
         dstPixel = srcPixel;
-      } else if (maskPixel < 1.0) {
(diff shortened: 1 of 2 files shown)


royshil:
> It works! (The latest commit)

not done yet
try setting to more than 1 frame calculation - it's flickering
but the fix is in

royshil:
@umireon this is ready for your review

src/background-filter.cpp
```diff
@@ -430,6 +430,28 @@ static void processImageForBackground(struct background_removal_filter *tf,
...
+      const float maskPixel = (float)(maskRow[j]) / 255.0f;
+      cv::Vec4b &dstPixel = dstRow[j];
+
+      if (maskPixel == 1.0) {
+        dstPixel = srcPixel;
+      } else if (maskPixel > 0.0) {
+        dstPixel = dstPixel * (1.0 - maskPixel) + srcPixel * (maskPixel);
+      }
+    }
+  }
```
```suggestion
  for (size_t i = 0; i < dst.total(); i++) {
    const auto &srcPixel = src.at<cv::Vec4b>(i);
    const auto maskPixel = mask.at<uchar>(i) / 255.0;
    auto &dstPixel = dst.at<cv::Vec4b>(i);

    dstPixel = dstPixel * (1.0 - maskPixel) + srcPixel * maskPixel;
  }
```

I think this way is more efficient for the C++ optimizer. Mat.at has no performance penalty when the build is Release.

src/background-filter.cpp
```diff
@@ -430,6 +430,22 @@ static void processImageForBackground(struct background_removal_filter *tf,
...
 
+// Blend
+void blend_images_with_mask(cv::Mat &dst, const cv::Mat &src, const cv::Mat &mask)
+{
+  for (size_t i = 0; i < dst.total(); i++) {
+    const auto &srcPixel = src.at<cv::Vec4b>(i);
+    const auto maskPixel = mask.at<uchar>(i) / 255.0f;
+    auto &dstPixel = dst.at<cv::Vec4b>(i);
+
+    if (maskPixel == 1.0f) {
```
royshil: @umireon i made an optimization here, since almost all pixels are either 1.0 or 0.0
only a few will be >0.0 && <1.0 - the feathering area
so we can skip the multiplication etc. in ~95% of the pixels

umireon: I think the if branch costs more than calculating all pixels but I'm not sure

umireon: I suppose that the compiler uses SIMD instructions with unrolling when there is no branch.

drieks:
Hi, I tried it (the build for this pull request) on Linux with OBS 29.0.2 and I'm seeing many messages per second
```
bad src image pointers
error: video_scaler_scale: sws_scale failed: -22
```

drieks:
It's also happening with the main branch build, so not related to this PR, sorry

umireon (approved):
LGTM!

PR #220, opened 2023-03-28
GS functions based rendering (supporting Browser sources and Groups)

umireon:
I have finally made rendering the portrait without the background using the OBS Graphics subsystem.
This way is more reliable and performant than the filter_video way.
At least, we don't anymore need to tweak obs_source_frame for transparent outputs.

This way adds support for Browser sources and Groups so that this change is also an enhancement.

There is one problem to adopt these changes: the multithreaded performance is so improved that frame drop would never happen and it increases CPU usage at most twice as much as the filter_video way.
It can be mitigated by setting calculate every X frame feature or stop calculating matte when the previous calculation has not been ended.


src/background-filter.cpp
```diff
@@ -506,37 +467,66 @@ static struct obs_source_frame *filter_render(void *data, struct obs_source_fram
...
+  gs_stage_texture(stagesurface, gs_texrender_get_texture(texrender));
+  uint8_t *video_data;
+  uint32_t linesize;
+  if (!gs_stagesurface_map(stagesurface, &video_data, &linesize)) {
+    return;
+  }
+  tf->inputBGRA = cv::Mat(height, width, CV_8UC4, video_data, linesize);
+  gs_stagesurface_unmap(stagesurface);
+  gs_stagesurface_destroy(stagesurface);
+  gs_texrender_destroy(texrender);
```
royshil, line 510: i wonder if these can be cached for subsequent frames and not built and destroyed every frame

royshil (approved):
very cool! well done @umireon ! 
i was experimenting with the graphics API. but this works perfect

i tested on Mac and windows, and seems to work great and improve performance

src/background-filter.cpp
umireon: I will change the code as you said in another PR.

PR #222, opened 2023-03-29
Use placement new and explicit destructor

umireon:
This ensures cv::Mat will be allocated and freed properly.

### DIFF
--- a/src/background-filter.cpp
+++ b/src/background-filter.cpp
@@ -56,7 +56,6 @@ struct background_removal_filter {
   std::vector<std::vector<int64_t>> outputDims;
   std::vector<std::vector<float>> outputTensorValues;
   std::vector<std::vector<float>> inputTensorValues;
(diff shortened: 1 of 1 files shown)


umireon:
OBS requires us to return void* pointer for data so we cannot use unique_ptr for tf.
Additionally, unique_ptr won't work correctly when the constructor and the destructor of the container struct are not called.
I don't think unique_ptrs on struct background_removal_filter are released when the filter is destroyed.
We have to call the constructor and the destructor of struct background_removal_filter.

royshil (approved):
OK i guess this is fine.

but are we releasing the  `bmalloc` somwhere?

umireon:
Releasing here and calling the destructor.
https://github.com/royshil/obs-backgroundremoval/blob/1cd54a9d28bde359b9beec8cceae0f02350e5224/src/background-filter.cpp#L320

PR #224, opened 2023-03-29
Fixes #223 (Add locks for cv::Mat and skip processing when the source is deactivated)

### LINKED ISSUE #223: Race condition against cv::Mat causes OBS to crash
**Describe the bug**
When changing certain settings, such as deactivating and re-activating a source, OBS Studio will crash.

**To Reproduce**
This bug seems to occur at random while changing settings, but this is how I reproduced it:
1. Add a Video Capture Device to any scene
2. Add a Background Removal filter
3. Deactivate and reactivate the video capture device
4. Crash

**Expected behavior**
OBS does not crash when using this plugin

**Screenshots**

 - OS: Windows 10 (22H2)
 - Browser: N/A
 - Plugin Version: f96d5da, as provided by GitHub Actions
 - OBS Version 29.0.2

**Additional context**
Crash reports:
[Crash 2023-03-29 15-40-19.txt](https://github.com/royshil/obs-backgroundremoval/files/11096637/Crash.2023-03-29.15-40-19.txt)

[Crash 2023-03-29 15-51-41.txt](https://github.com/royshil/obs-backgroundremoval/files/11096694/Crash.2023-03-29.15-51-41.txt)

Logs:
[2023-03-29 15-41-50.txt](https://github.com/royshil/obs-backgroundremoval/files/11096640/2023-03-29.15-41-50.txt)

[2023-03-29 15-46-58.txt](https://github.com/royshil/obs-backgroundremoval/files/11096693/2023-03-29.15-46-58.txt)

umireon:
Prevent race condition from happening.

Hopefully fixes #223

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,9 +112,6 @@ if(OS_WINDOWS)
   if(MSVC)
     target_compile_options(${CMAKE_PROJECT_NAME} PRIVATE /W4)
   endif()
-
-  target_include_directories(${CMAKE_PROJECT_NAME} SYSTEM
-                             PRIVATE ${CMAKE_SOURCE_DIR}/../obs-studio/deps/w32-pthreads)
   # --- End of section ---
 
   # -- macOS specific build settings and tasks --
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,6 +112,9 @@ if(OS_WINDOWS)
   if(MSVC)
     target_compile_options(${CMAKE_PROJECT_NAME} PRIVATE /W4)
   endif()
+
+  target_include_directories(${CMAKE_PROJECT_NAME} SYSTEM
+                             PRIVATE ${CMAKE_SOURCE_DIR}/../obs-studio/deps/w32-pthreads)
   # --- End of section ---
 
   # -- macOS specific build settings and tasks --
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,7 +113,7 @@ if(OS_WINDOWS)
     target_compile_options(${CMAKE_PROJECT_NAME} PRIVATE /W4)
   endif()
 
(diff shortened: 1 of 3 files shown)


src/background-filter.cpp
```diff
@@ -76,6 +77,9 @@ struct background_removal_filter {
   cv::Mat inputBGRA;
   cv::Mat outputBGRA;
 
+  pthread_mutex_t inputBGRALock;
+  pthread_mutex_t outputBGRALock;
```
royshil: i think we should use `std`'s locking mechanism

e.g. use `std::mutex` (https://en.cppreference.com/w/cpp/thread/mutex) and a `scoped_lock` (https://en.cppreference.com/w/cpp/thread/scoped_lock)

it's much more idiomatic. i don't like `pthreads` as an added dependency either
should make the code cleaner too...

royshil (changes requested):
let's use `std` tools for locking

umireon:
I've changed the code to work with std::mutex but it revealed that using std::mutex requires the constructor of background_removal_filter to be called and thus requires https://github.com/royshil/obs-backgroundremoval/pull/222 to be merged.

CMakeLists.txt
```diff
@@ -112,6 +112,9 @@ if(OS_WINDOWS)
   if(MSVC)
     target_compile_options(${CMAKE_PROJECT_NAME} PRIVATE /W4)
   endif()
+
+  target_include_directories(${CMAKE_PROJECT_NAME} SYSTEM
+                             PRIVATE ${CMAKE_SOURCE_DIR}/../obs-studio/deps/w32-pthreads)
```
royshil: we don't need this now right? @umireon

umireon: Removed

src/background-filter.cpp
umireon: Adopted std::mutex

royshil (approved):
this is looking good

PR #225, opened 2023-03-29
Use uint8_t buffer for inter-thread communication

umireon:
Referencing cv::Mat from two different threads may cause an OBS crash.
https://github.com/royshil/obs-backgroundremoval/issues/223

By this change, all the cv::Mat will be referenced by only a single thread and may suppress crashes on cv::Mat.clone()


royshil (changes requested):
i'm very much against this approach.
it makes the code all that much unreadable, an we carry all these pointers and members around...

really, i don't think we should do it this way.

umireon:
We are going with the mutex approach!

PR #227, opened 2023-03-29
Fix escaping on postinstall of Mac pkg

umireon:
The installer has failed to install when the path of pkg contains any spaces such as `obs-backgroundremoval-0.5.15-macos-arm64 (1).pkg`.
This change escapes spaces properly and fixes the installation problem on Mac.


royshil (approved):
perfect. good idea

PR #229, opened 2023-03-29
Provide only the universal binary for Mac

umireon:
So many people have tried to install ARM64 binary with Intel OBS.
The size of the universal installer is almost the same as the ARM64 and the x86_64 ones so I can say that there is no disadvantage to quitting providing the other installers.
By providing only the universal installer, the number of people who report "Simply not working" would decrease.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -266,18 +266,18 @@ jobs:
           cp build_x86_64/installer-macos.generated.pkgproj build_arm64/
           cp build_x86_64/installer-macos.generated.pkgproj build_universal/
 
-      - name: Package x86_64 Plugin
-        uses: ./plugin/.github/actions/package-plugin
-        with:
-          workingDirectory: ${{ github.workspace }}/plugin
-          target: x86_64
-          config: RelWithDebInfo
-          codesign: ${{ github.event_name != 'pull_request' && steps.setup.outputs.haveCodesignIdent == 'true' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -266,18 +266,18 @@ jobs:
...
-          workingDirectory: ${{ github.workspace }}/plugin
-          target: x86_64
-          config: RelWithDebInfo
-          codesign: ${{ github.event_name != 'pull_request' && steps.setup.outputs.haveCodesignIdent == 'true' }}
-          notarize: ${{ startsWith(github.ref, 'refs/tags/') && github.event_name != 'pull_request' && steps.setup.outputs.haveNotarizationUser == 'true' }}
-          codesignIdent: ${{ env.CODESIGN_IDENT }}
-          installerIdent: ${{ env.CODESIGN_IDENT_INSTALLER }}
-          codesignUser: ${{ secrets.MACOS_NOTARIZATION_USERNAME }}
-          codesignPass: ${{ secrets.MACOS_NOTARIZATION_PASSWORD }}
+      # - name: Package x86_64 Plugin
```
royshil: instead of commenting out we can just use the `-if: false` clause to prevent this from running

royshil (changes requested):
please explore the `if` functionality in the GH actions

.github/workflows/main.yml
umireon: Fixed

royshil (approved):
perfect!
now we can bring it back if we want in the future

PR #233, opened 2023-03-30
Execute obs_source_skip_video_filter when vider_render has no output

umireon:
This would improve the performance a bit.


royshil (approved):
looks good!

Issue #182, opened 2023-03-16
GCC build on Fedora and Arch is broken


umireon:
Due to https://github.com/microsoft/onnxruntime/issues/14923, the builds using GCC fail with some warnings.
We have to update the ONNX Runtime after this issue is resolved and there is almost nothing to fix this on our side.
This breaks the flatpak build, Fedora build, and Arch build.
Related #132 #172

drieks:
As a workaround, I  added 
```
#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
```
as first line here: onnxruntime/core/mlas/lib/intrinsics/avx512/quantize_avx512f.cpp

ngoonee:
Is my understanding correct that there's nothing to be done from obs-backgroundremoval side on this issue, and it is purely an upstream issue which can be worked around?

umireon:
Resolved by #250

PR #235, opened 2023-03-31
Use patched onnxruntime for Arch and Fedora build

umireon:
https://github.com/royshil/obs-backgroundremoval/issues/182


royshil:
@umireon I would very very much prefer if we didn't do it this way with an external dependent repo

If we need to patch onnxruntime we can bring the .patch diff file to this repo
Download the original source tar.gz and apply the patch file before building it.

umireon:
Upstream doesn't seem to fix the problem very soon so we need a patch for now, I suppose.

royshil (approved):
Looking good
Ship it!

PR #237, opened 2023-03-31
Do not execute video_tick when disabled to reduce CPU usage

### LINKED ISSUE #236: OBS Background Removal consumes excessive resources when effect is not visible
**Describe the bug**
When the background removal effect is not visible (hidden using the "eye" icon, or the source is deactivated), the background removal plugin uses approximately the same amount of CPU as when it is enabled and visible.

**To Reproduce**
1. Open Task Manager to the Performance tab to see the difference in resource usage
2. Open OBS, add a Background Removal filter to any source
3. Toggle the visibility of the effect (or deactivate/reactivate the source)
4. Notice that the CPU (and GPU, if using GPU inference) resources do not change until the effect is removed

**Expected behavior**
CPU/GPU resources drop when the plugin effect is disabled (but still in the list of effects for a source)

**Screenshots**
Annotated screenshot of the resource usage issue:
Capture2

 - OS: Windows 10 (22H2)
 - Browser: N/A
 - Plugin Version: Latest commit (f4f616c) from GitHub Actions
 - OBS Version: 29.0.2

**Additional context**
Log file: [2023-03-31 15-42-10.txt](https://github.com/royshil/obs-backgroundremoval/files/11118252/2023-03-31.15-42-10.txt)

umireon:
Fixes #236

### DIFF
--- a/src/background-filter.cpp
+++ b/src/background-filter.cpp
@@ -427,6 +427,15 @@ void filter_video_tick(void *data, float seconds)
 {
   struct background_removal_filter *tf = reinterpret_cast<background_removal_filter *>(data);
 
+  if (!obs_source_enabled(tf->source)) {
+    return;
+  }
+
+  obs_source_t *parent = obs_filter_get_parent(tf->source);
+  if (!parent) {
+    return;
+  }
+
   if (tf->inputBGRA.empty()) {
     return;
   }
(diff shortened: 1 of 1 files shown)


umireon:
It seems that we have to implement activate and deactivate

royshil (approved):
nice!

PR #241, opened 2023-04-01
Enable this filter when it is added to the source

umireon:
isDisabled is false when the filter is added.
It seems that activate is not called when the filter is added


royshil (approved):
This is a semantic change
Looks good

PR #244, opened 2023-04-05
Respect the output of the previous filter if exists

### LINKED ISSUE #243: OBS Filter Crop/Pad is not working properly when Background Removal is on
**Describe the bug**

OBS Filter Crop/Pad is not working properly when Background Removal is on

**To Reproduce**

1. Add Video Capture with a camera
2. Select the Filter from the Video Capture Source
3. Select Background Removal
4. Select the Blur Background factor as 1
5. Close
6. Select the Filter again from the Video Capture Source
7. Select Crop/Pad
8. Adjust Top as 300 and Bottom as 700
9. Compare when Background Removal is Off and On
10. The Top value in Crop/Pad does not work. (It crops from the bottom always)

**Expected behavior**

The Top Value in Crop/Pad should work as expected no matter what.

**Screenshots**

* When Background Removal is OFF
Background Removal Off

* When Background Removal is ON
Background Removal On

**Log and Crash Report**

[2023-04-03 20-22-30.txt](https://github.com/royshil/obs-backgroundremoval/files/11144415/2023-04-03.20-22-30.txt)

 - OS: macOS Ventura
 - Browser: Safari
 - Plugin Version: 0.5.16
 - OBS Version: 29.0.2

umireon:
Try to fix #243 

We have to use obs_filter_get_target to get the output of the previous filter and we have to use obs_source_get_base_width instead of obs_source_get_width

### DIFF
--- a/src/background-filter.cpp
+++ b/src/background-filter.cpp
@@ -530,13 +530,8 @@ static void filter_video_render(void *data, gs_effect_t *_effect)
     return;
   }
 
-  obs_source_t *parent = obs_filter_get_parent(tf->source);
-  if (!parent) {
-    obs_source_skip_video_filter(tf->source);
-    return;
-  }
-  const uint32_t width = obs_source_get_width(parent);
-  const uint32_t height = obs_source_get_height(parent);
+  const uint32_t width = obs_source_get_width(tf->source);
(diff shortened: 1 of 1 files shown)


royshil (approved):
Looking good!

PR #245, opened 2023-04-06
CUDA support on Linux

umireon:


umireon:
I concluded that enabling CUDA with linking ONNX Runtime statically was impossible.

PR #246, opened 2023-04-06
Adopt dynamically linked ONNX Runtime on macOS

umireon:
I have investigated how we can implement the matte on CUDA and concluded that we need to use the pre-built ONNX Runtime binaries provided by Microsoft and link ONNX Runtime dynamically because there is no way to install cuDNN which ONNX Runtime requires to be built on GitHub Actions.

To deliver CUDA implementation to Linux and Windows, I propose to link ONNX Runtime dynamically.

As the first trial, I want to dynamically link ONNX Runtime on macOS and include the dylib of ONNX Runtime to pkg.
It is a side effect but the build duration gets much shorter.


umireon:
This change won't require users to install ONNX Runtime separately because it is included in Contents/Frameworks directory of the plugin bundle.

umireon:
Review ready!

royshil (approved):
alright!
i checked it - works perfect
we can ship

this is in fact how i've had it in the past, e.g. not building ORT from scratch rather downloading it. it's much faster...

we should do this for Windows now as well!

PR #249, opened 2023-04-11
Use prebuilt library on Windows

umireon:


royshil (approved):
nice! let's land this

PR #250, opened 2023-04-12
Shared ONNX Runtime on Linux

umireon:
It seems that this configuration automatically loads CUDA and TensorRT EP when the CUDA library is available on the user's system.


royshil (approved):
nice - let's ship

Issue #242, opened 2023-04-03
CUDA support on linux


grinco:
I installed the plugin using flatpak: https://github.com/flathub/flathub/pull/4016 and only CPU support was available.

The CPU utilization while using the plugin on a 9th gen i7 CPU is 400% (uses half of all available cores at all time). Ideally, we would like to see hardware offloading. This issue is specifically requesting CUDA support on linux.

The plugin has been tested on the following: NVIDIA-SMI 525.105.17 Driver Version: 525.105.17 CUDA Version: 12.0

umireon:
I've started to implement CUDA to this plugin #246.

umireon:
I've concluded that GPU support on Flatpak is not feasible.

umireon:
I've implemented a testing GPU support on Linux. Can you test this deb file?
This requires TensorRT 8.5, CUDA 11.6, and cuDNN 8.2.4. I think the versions must be matched exactly.

https://github.com/royshil/obs-backgroundremoval/suites/12186836154/artifacts/643601399

grinco:
I'm using an rpm based system. Do you happen to have an RPM, too? If suggest using build.opensuse.org - they have a great build system that can produce binaries for various os and architectures. Really great stuff. If you point me to the commit or branch you built from I can build it locally, too.

grinco:
Nevermind. I see the pull request reference. Will give it a try.

umireon:
I'm not sure it works because I don't have Linux with GPU/

lgeekjopt:
So i downloaded your zip. And installed your deb
Installed on Mint 21.1 (Ubuntu 22.04). I use obs from ppa. 

"NVIDIA-SMI 510.108.03   Driver Version: 510.108.03   CUDA Version: 11.6"

error: [obs-backgroundremoval] /onnxruntime_src/onnxruntime/core/session/provider_bridge_ort.cc:1106 onnxruntime::Provider& onnxruntime::ProviderLibrary::Get() [ONNXRuntimeError] : 1 : FAIL : Failed to load library libonnxruntime_providers_tensorrt.so with error: libcudnn.so.8:

grinco:
@umireon so after compiling it manually for opensuse tumbleweed, I was able to run it, however, I'm getting a lookup error:
obs: symbol lookup error: /usr//lib64/obs-plugins/libonnxruntime_providers_cuda.so: undefined symbol: Provider_GetHost  

[...]

PR #257, opened 2023-04-12
Fix symbol not found Providers_GetHost

umireon:


umireon:
I'm asking for testers on #242

grinco:
Can confirm the symbol error is gone, however video freezes once I switch to tensorrt. I am suspecting that it's due to my local set up - where tensort rt, and cudnn are installed in a miniconda env. Someone with a supported OS must confirm.

umireon:
@grinco Can you attach here the OBS log?

umireon:
@grinco reported that the problem which this PR wants to fix is fixed.

> Can confirm the symbol error is gone

This PR blocks #266 so I'd like to merge this PR and handle remaining problems that @grinco reported on #242.

@royshil Can you approve this PR?

royshil (approved):
lgtm

PR #275, opened 2023-04-24
Add Default settings to ORT sess

royshil:
on windows default to DML, mac to CoreML
new users should not have to manually select these options after the fact, we should attempt to use them since they provide a great performance boost


umireon:
I'm afraid that on Mac CPU matte is more efficient than CoreML so we might want to keep the Mac default CPU?

royshil:
@umireon i've changed it back to CPU for Mac

umireon (approved):
LGTM!

PR #276, opened 2023-04-24
Use Kawase Blur

### LINKED ISSUE #271: Blur Background setting causing significant GPU usage above certain settings
**Describe the bug**

When setting the `Blur background factor` setting high enough, the background blur effect will use a very significant amount of GPU resources (this is more apparent on integrated graphics).

**To Reproduce**

1. Open Task Manager and OBS Studio
2. Add a `Background Removal` filter to any source (if you haven't already).
3. Set `Blur Background Factor` to a high setting (settings as low as 6 max out my integrated graphics, while my discrete graphics can handle up to around 12 before lagging significantly)
4. Observe (in Task Manager) the High GPU usage

**Expected behavior**

The background blur does not use this much GPU time.

**Screenshots**

Before (tested using 39adf39) | After (latest release)
------------------------------|------------------------
before | after

Background blur factor was set to 50 (before) and 12 (after). These settings give about the same amount of blur when switching versions, from my testing.

**Log and Crash Report**

[2023-04-24 07-53-54.txt](https://github.com/royshil/obs-backgroundremoval/files/11305075/2023-04-24.07-53-54.txt)

[2023-04-24 07-55-51.txt](https://github.com/royshil/obs-backgroundremoval/files/11305074/2023-04-24.07-55-51.txt)

 - OS: Windows 10 (22H2)
 - Browser: N/A
 - Plugin Version: Latest (v0.5.17)
 - OBS Version: Latest (v29.0.2)

**Additional context**

Commit 39adf39 and earlier do not seem to exhibit this behaviour, which suggests that the newly-introduced GPU blending in the latest release is actually performing worse than before

umireon:
Fixes #271

https://www.intel.com/content/www/us/en/developer/articles/technical/an-investigation-of-fast-real-time-gpu-based-image-blur-algorithms.html#:~:text=%C2%A0%22Kawase%20blur%22%20filter

### DIFF
--- a/src/background-filter.cpp
+++ b/src/background-filter.cpp
@@ -520,9 +520,17 @@ static gs_texture_t *blur_background(struct background_removal_filter *tf, uint3
     gs_effect_set_texture(image, blurredTexture);
     gs_effect_set_float(xOffset, (i + 0.5f) / width);
     gs_effect_set_float(yOffset, (i + 0.5f) / height);
+
+    struct vec4 background;
+    vec4_zero(&background);
+    gs_clear(GS_CLEAR_COLOR, &background, 0.0f, 0);
+    gs_ortho(0.0f, static_cast<float>(width), 0.0f, static_cast<float>(height), -100.0f, 100.0f);
+    gs_blend_state_push();
+    gs_blend_function(GS_BLEND_ONE, GS_BLEND_ZERO);
     while (gs_effect_loop(tf->kawaseBlurEffect, "Draw")) {
       gs_draw_sprite(blurredTexture, 0, width, height);
     }
+    gs_blend_state_pop();
     gs_texrender_end(tf->texrender);
     gs_copy_texture(blurredTexture, gs_texrender_get_texture(tf->texrender));
   }
--- a/src/background-filter.cpp
+++ b/src/background-filter.cpp
@@ -634,6 +634,7 @@ static void filter_video_render(void *data, gs_effect_t *_effect)
   gs_blend_state_pop();
 
   gs_texture_destroy(alphaTexture);
+  gs_texture_destroy(blurredTexture);
 }
 
 struct obs_source_info background_removal_filter_info = {
--- a/src/background-filter.cpp
+++ b/src/background-filter.cpp
@@ -501,18 +501,15 @@ void filter_video_tick(void *data, float seconds)
 static gs_texture_t *blur_background(struct background_removal_filter *tf, uint32_t width,
                                      uint32_t height)
 {
-  if (tf->blurBackground == 0.0) {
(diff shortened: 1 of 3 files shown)


src/background-filter.cpp
```diff
@@ -554,6 +561,30 @@ static void filter_video_render(void *data, gs_effect_t *_effect)
 
   // Output the masked image
 
+  gs_texture_t *blurredTexture = gs_texture_create(width, height, GS_BGRA, 1, nullptr, 0);
+  gs_copy_texture(blurredTexture, gs_texrender_get_texture(tf->texrender));
+  for (int i = 0; i < (int)tf->blurBackground; i++) {
```
royshil: we can move many things out of the loop...

src/background-filter.cpp
```diff
@@ -554,6 +561,30 @@ static void filter_video_render(void *data, gs_effect_t *_effect)
 
   // Output the masked image
 
+  gs_texture_t *blurredTexture = gs_texture_create(width, height, GS_BGRA, 1, nullptr, 0);
```
royshil: we should only perform any of this if `tf->backgroundBlur > 0`

royshil (changes requested):
this looks good - but lets clean it up first

src/background-filter.cpp
umireon: Fixed

src/background-filter.cpp
umireon: Fixed!

royshil (approved):
nice! well done! go

royshil:
@umireon  wait! we got a problem
works great in the filters screen
image
but in the main view....
image
we have a problem

umireon:
On my side, everything works fine.

スクリーンショット 2023-04-25 23 31 21
スクリーンショット 2023-04-25 23 31 18

royshil:
@umireon i'm still having a problem, and getting a strange error on the log
```
14:26:30.848: device_texture_create (D3D11): Failed to create 2D texture (80070057)
14:26:30.848: [obs-backgroundremoval] Failed to create alpha texture
14:26:30.853: device_texture_create (D3D11): Failed to create 2D texture (80070057)
14:26:30.853: [obs-backgroundremoval] Failed to create alpha texture
```
this error doesn't come up on the main version of the plugin.

so i think it's related to this change you're proposing here

umireon:
@royshil Can you paste the full log here?

umireon:
I could reproduced the problem on my side

スクリーンショット 2023-04-26 13 40 57

umireon:
Fixed

royshil (approved):
confirmed to fix the problem - fire at will! 🚀

PR #277, opened 2023-04-25
Avoid using DML deprecated API

umireon:
> * [[deprecated]]
> * This export is deprecated.
> * The OrtSessionOptionsAppendExecutionProvider_DML export on the OrtDmlApi should be used instead.

https://github.com/microsoft/onnxruntime/blob/main/include/onnxruntime/core/providers/dml/dml_provider_factory.h#L33-L43


royshil (approved):
this works for me on windows, i think it's OK to merge

PR #278, opened 2023-04-25
Arch CI

umireon:
This change enables GitHub Actions to check if PKGBUILD can be built.


royshil (approved):
very cool

but - do you want to exatrct the built package from there and offer it for download?

umireon:
I suppose we don't need to distribute the package because it is available on AUR.
https://aur.archlinux.org/packages/obs-backgroundremoval

PR #284, opened 2023-04-27
Add the link to AUR

umireon:
It is enough for Arch Linux users to mention AUR.
Each Arch user builds an AUR package in their own way.


royshil (approved):
go

PR #297, opened 2023-05-02
Adding enhance image net

royshil:
- Restructuring our internal APIs
- Adding a new filter for low-light image enhancement based on TBEFN


umireon:
Do we need to add an attribution for the original author of the model?

umireon (approved):
Looks great and worked on my machine!

PR #300, opened 2023-05-05
Add GPU to enhance filter

royshil:
Adding the GPU selection options, as well as adding high pass filter to enhancement output which is blurry (because the networks run on a lower resolution).


umireon (approved):
It works!

PR #309, opened 2023-05-20
Make it easier to spot the openSUSE installation instructions

aspiers:
The very helpful openSUSE installation instructions were not referenced or linked from anywhere, so they were easy to miss.

So link to them from the top-level `README.md`.


umireon (approved):
Nice catch!

Issue #315, opened 2023-05-31
Better matting capabilities of e.g. Hair and Glasses


royshil:
**Is your feature request related to a problem? Please describe.**

I would like to get a more accurate matting (separation between FG and BG), not just a "binary blob" of FG.
I've seen Google Meet be very good with matting e.g. hair and glasses.

**Describe the solution you'd like**

A better matting of partial foregrounds, e.g hair

**Describe alternatives you've considered**

Currently we have a "blob", which we can apply feathering to.

**Additional context**

N/A

umireon:
Sounds great!

royshil:
this was largely implemented by #320

PR #320, opened 2023-06-02
Add "no threshold" options

royshil:
This adds the ability to disable thresholding all together 
It saves greatly on CPU, and allows for a more "soft" segmentation (e.g. with hair or glasses) #315


umireon (approved):
LGTM

PR #330, opened 2023-06-11
We have to state that the pre-built binary only works on Ubuntu 22.04 or later

umireon:
On Ubuntu 20.04, libstdc++ version won't match to the one this plugin expects and fail to load with unfriendly message.
We have to state that only Ubuntu 22.04 or later is supported by our pre-built binaries.


royshil (approved):
this makes sense. we can add this to the readme

PR #337, opened 2023-06-18
Require OBS 28+ on the deb package

umireon:
This plugin actually needs OBS 28+.
Qt dependencies are built in the main OBS package so I want to omit them.


royshil (approved):
nice catch! 👏

PR #339, opened 2023-06-19
Remove zip from releases page

umireon:
Many people on Windows are confused about whether they should use zip or exe and end up with a broken installation.
I suggest removing zip artifacts from the releases page and checking if there are people who want the zip archive.


royshil:
i don't know about this... i think the .zip option is still useful for advanced users.
maybe we can add more instruction for users to make them pick the default option?
or maybe we can add a ADVANCED as a prefix for the files? 🤷‍♂️  or somehow separate them
what do you think

umireon:
How about suffixing zip with 'portable'?

royshil:
Yes this is good.

royshil (approved):
LGTM

PR #341, opened 2023-06-19
Release unofficial a Fedora RPM for obs-stufio on RPM Fusion

umireon:
Obs-studio is distributed on RPM Fusion so I think it is reasonable to ship our plugin with RPM.
We should ship our RPM as an unofficial or experimental distribution because the OBS project ships their obs-studio RPM as unofficial distribution.


.github/workflows/main.yml
```diff
@@ -494,6 +494,39 @@ jobs:
...
+      image: "fedora:37"
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        arch: [x86_64]
+    if: always()
+    needs: [clang_check]
+    steps:
+      - name: "Create an user"
```
royshil: nit: `Create a user`

royshil (approved):
LGTM
fix n ship

.github/workflows/main.yml
umireon: Fixed! I'm bad at writing English 🤣

umireon:
@royshil Can you add documentation about this RPM after this PR is merged? Or do you want me to add a note?

Issue #342, opened 2023-06-19
Add this plugin to Packman


umireon:
Packman is where the unofficial OBS binary for OpenSUSE is hosted.

https://pmbs.links2linux.org/

- [ ] Register to Packman
- [ ] Create the definition

royshil:
@umireon what's the status of this? anything i can help with?

umireon:
I asked on the mailing list of Packman but I've got no response. I will ask the list again later.

royshil:
let's get this off of milestone v1.1.0 and on to v1.1.1 then

umireon:
#425 needs to be merged.

umireon:
I have submitted the request to add our plugin to the repository again.

PR #343, opened 2023-06-22
Allow users to use a custom ONNX Runtime tarball

umireon:
This is useful for Flatpak and Open Build Service (#342).

### DIFF
--- a/cmake/FetchOnnxruntime.cmake
+++ b/cmake/FetchOnnxruntime.cmake
@@ -1,11 +1,27 @@
 include(FetchContent)
 
+set(CUSTOM_ONNXRUNTIME_URL
+    ""
+    CACHE STRING "URL of a downloaded ONNX Runtime tarball")
+
+set(CUSTOM_ONNXRUNTIME_MD5
+    ""
+    CACHE STRING "MD5 Hash of a downloaded ONNX Runtime tarball")
+
 set(Onnxruntime_VERSION "1.15.1")
+
 if(OS_MACOS)
-  FetchContent_Declare(
-    Onnxruntime
-    URL "https://github.com/microsoft/onnxruntime/releases/download/v${Onnxruntime_VERSION}/onnxruntime-osx-universal2-${Onnxruntime_VERSION}.tgz"
-    URL_HASH MD5=6637c31d2dae5ad8bffb4da38ae4d89e)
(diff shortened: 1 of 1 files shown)


cmake/FetchOnnxruntime.cmake
```diff
@@ -1,11 +1,27 @@
...
-    Onnxruntime
-    URL "https://github.com/microsoft/onnxruntime/releases/download/v${Onnxruntime_VERSION}/onnxruntime-osx-universal2-${Onnxruntime_VERSION}.tgz"
-    URL_HASH MD5=6637c31d2dae5ad8bffb4da38ae4d89e)
+  if(CUSTOM_ONNXRUNTIME_URL STREQUAL "")
+    FetchContent_Declare(
+      Onnxruntime
+      URL "https://github.com/microsoft/onnxruntime/releases/download/v${Onnxruntime_VERSION}/onnxruntime-osx-universal2-${Onnxruntime_VERSION}.tgz"
+      URL_HASH MD5=6637c31d2dae5ad8bffb4da38ae4d89e)
+  else()
+    FetchContent_Declare(
```
royshil, line 31: can you please check that if URL is not empty then MD5 must not be empty?

cmake/FetchOnnxruntime.cmake
```diff
@@ -1,11 +1,27 @@
...
+    CACHE STRING "MD5 Hash of a downloaded ONNX Runtime tarball")
+
 set(Onnxruntime_VERSION "1.15.1")
+
 if(OS_MACOS)
-  FetchContent_Declare(
-    Onnxruntime
-    URL "https://github.com/microsoft/onnxruntime/releases/download/v${Onnxruntime_VERSION}/onnxruntime-osx-universal2-${Onnxruntime_VERSION}.tgz"
-    URL_HASH MD5=6637c31d2dae5ad8bffb4da38ae4d89e)
+  if(CUSTOM_ONNXRUNTIME_URL STREQUAL "")
```
royshil: so instead of having multiple `CUSTOM_ONNXRUNTIME_URL STREQUAL ""` just have a boolean for using the custom URL and use that instead in all places, and you can check the MD5 there as well

royshil (approved):
looking good! fix and ship

cmake/FetchOnnxruntime.cmake
umireon: Added validation

PR #349, opened 2023-06-24
Support providing a custom OpenCV tarball

umireon:


cmake/BuildMyOpenCV.cmake
```diff
@@ -1,5 +1,26 @@
 include(ExternalProject)
 
+set(CUSTOM_OPENCV_URL
+    ""
+    CACHE STRING "URL of a downloaded OpenCV tarball")
```
royshil: Nit: you don't need "downloaded"
Just "URL of OpenCV tarball" can work ok

cmake/BuildMyOpenCV.cmake
```diff
@@ -1,5 +1,26 @@
 include(ExternalProject)
 
+set(CUSTOM_OPENCV_URL
+    ""
```
royshil: By the way if you know the default value (below, from GitHub) why not use it here as default value?

cmake/BuildMyOpenCV.cmake
umireon: Fixed!

cmake/BuildMyOpenCV.cmake
umireon: Fixed!

royshil (approved):
Looking good

PR #348, opened 2023-06-24
Fix to include libonnxruntime.1.15.1

umireon:
Closes #347

PR #350, opened 2023-06-24
Link libonnxruntime.so with absolute path when NOT LINUX_PORTABLE

umireon:
It would be great if $ORIGIN is not included in libraries when building rpm.
This option enable builders to build this plugin without $ORIGIN.


royshil (approved):
This checks out

umireon:
I can check this works after #348 is merged

umireon:
It revealed that this fix won't mitigate the rpmlint error so I will close this PR.

PR #357, opened 2023-06-28
Update our build system with obs-plugintemplate

umireon:
The build system on https://github.com/obsproject/obs-plugintemplate was updated.
It would be great to update our build system also.


umireon:
I confirmed the notarization works.

buildspec.json
```diff
@@ -1,86 +1,57 @@
...
-            "version": "28.0.1",
-            "repository": "https://github.com/obsproject/obs-studio.git",
-            "branch": "master",
-            "hash": "e8dc70d0eef4503378d6ac300e680215eb5c9379"
-        },
-        "prebuilt": {
-            "version": "2022-08-02",
-            "baseUrl": "https://github.com/obsproject/obs-deps/releases/download",
-            "label": "Pre-built obs-deps",
+            "version": "29.1.2",
```
royshil: isnt this going to limit us to run on OBS v29.1.2 and up?
i think we should stick with v28 if possible

src/plugin-support.h
```diff
@@ -1,6 +1,6 @@
 /*
-@PLUGIN_PRETTY_NAME@
-Copyright (C) 2021 @PLUGIN_AUTHOR@ @LINUX_MAINTAINER_EMAIL@
+OBS Background Removal Filter Plugin
+Copyright (C) 2021 Roy Shilkrot roy.shil@gmail.com
```
royshil, line 3: shuold we use the variables from buildpec,json

royshil (approved):
LGTM but check out a couple comments

buildspec.json
umireon: I forgot to fix this.

umireon: Fixed

src/plugin-support.h
umireon: Fixed

buildspec.json
umireon: It revealed that the new build system does not support OBS 28.0.1.

royshil: i see it fails on mac because of SWIG
can we install SWIG to resolve?

umireon: I have modified the build system and managed to be built on my local Mac with OBS 28.0.1.

umireon: I conclude that the new build system requires huge modification if we go with OBS 28.0.1 and this is not reasonable.

umireon: I tested the configuration of OBS 28.0.1 and so many problems other than SWIG occurred.

royshil:
> I have to say that we have to bump OBS dependency to 29+ to adopt the new build system. The problem that occurs when we use OBS 28.0.1 is not so simple.

what do you suggest we do?
we update our support to 29.1+? 
many users are on 28, especially in Linux

umireon:
@royshil Finally I have this plugin built with OBS 28.0.1.

royshil (approved):
alright looking good!

PR #359, opened 2023-06-30
Compile with QNetworkAccessManager

umireon:


umireon:
It is impossble to combine Qt6 with OBS 28.

PR #360, opened 2023-06-30
We need to update the OBS dependency to 29.

umireon:
I finally finally concluded that we have to use OBS 29 with the new build system and Qt6.
It is nearly impossible to build both with Qt6 and with OBS 28.

See https://github.com/royshil/obs-backgroundremoval/pull/359

I think our plugin may work with OBS 28 when the build dependency is OBS 29.

### DIFF
--- a/buildspec.json
+++ b/buildspec.json
@@ -1,33 +1,33 @@
 {
     "dependencies": {
         "obs-studio": {
-            "version": "28.0.1",
+            "version": "29.1.2",
             "baseUrl": "https://github.com/obsproject/obs-studio/archive/refs/tags",
             "label": "OBS sources",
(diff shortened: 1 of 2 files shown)


umireon:
I cannot install OBS 28 on my machine because all the OBS distribution is already OBS 29.
I think this means OBS 28 is already legacy and we have to advise users to update to OBS 29+.

umireon:
I cannot spare any more time to investigate to make this plugin buildable with OBS 28 binaries.

umireon:
We can still say that OBS 28 users must use older versions of this plugin.

royshil (approved):
OK!
no problem. we have older versions that work with v28

PR #362, opened 2023-06-30
Compile with QNetworkAccessManager 2

umireon:
This change will link the networking library of Qt to the plugin.
To check if QtNetwork is linked, I added QNetworkAccessManager object.


umireon:
I had believed that the compilation error was caused by OBS 28 but it revealed not to be true

royshil (approved):
looks great!

PR #363, opened 2023-07-01
Fix Ubuntu build

umireon:
This change will suppress the dpkg-shlibdeps errors.
https://github.com/royshil/obs-backgroundremoval/actions/runs/5427808368/jobs/9871413794#step:7:234


royshil (approved):
ship it!

PR #364, opened 2023-07-01
An update dialog skelton

umireon:
スクリーンショット 2023-07-01 16 17 11

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,6 +73,8 @@ target_sources(
   PRIVATE src/plugin-main.c
           src/ort-utils/ort-session-utils.cpp
           src/obs-utils/obs-utils.cpp
+          src/update-checker/update-checker.cpp
+          src/update-checker/UpdateDialog.cpp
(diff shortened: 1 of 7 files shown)


royshil (approved):
i'm loving it
but we shouldn't show this to the users right now
so we can keep it here on `main`, just don't add it to the plugin startup
we will build the version checker and the dialog itself, and then we can add it the startup

umireon:
@royshil Can you create a new topic branch for a version checker? I will merge this into that branch.

royshil:
> @royshil Can you create a new topic branch for a version checker? I will merge this into that branch.

you can certainly merge all of this on `main` for now, but we will connect the plugin startup code in a bit after we bring in the other libcurl part

umireon:
@royshil I have removed the code of this PR from plugin-main.c. Is this what you mean?

royshil:
> @royshil I have removed the code of this PR from plugin-main.c. Is this what you mean?

@umireon yes. we're good now!

PR #365, opened 2023-07-01
Add libcurl to fetch GitHub metadata

umireon:
I tried QtNetwork but it needs OpenSSL to connect to HTTPS endpoints and obs-deps do not include OpenSSL.
libcurl can utilize various types of SSL implementation and it can connect to HTTPS endpoints without adding any heavy dependencies to this plugin.


umireon:
I have confirmed that this works on Windows, Linux, and macOS

CMakeLists.txt
```diff
@@ -68,11 +68,38 @@ else()
   target_link_libraries(${CMAKE_PROJECT_NAME} PRIVATE OpenCV)
 endif()
 
+set(MbedTLS_DIR ${CMAKE_PREFIX_PATH})
```
royshil: let's get all of this to an extrnal file in `cmake/`
like `BuildibCurl.cmake` or something
and the `include()` it here

CMakeLists.txt
```diff
@@ -68,11 +68,38 @@ else()
...
+  add_compile_options(-fPIC)
+  set(CURL_USE_OPENSSL ON)
+  set(CURL_USE_NSS OFF)
+endif()
+set(BUILD_CURL_EXE OFF)
+set(BUILD_SHARED_LIBS OFF)
+set(HTTP_ONLY ON)
+set(HTTP_ONLY OFF)
+set(CURL_USE_LIBSSH2 OFF)
+add_subdirectory(vendor/curl EXCLUDE_FROM_ALL)
```
royshil: why submodule and not what we do with OpenCV? e.g. the CMake external build

build-aux/run-cmake-format
```diff
@@ -70,6 +70,7 @@ invoke_formatter() {
 
       local -a source_files=(**/(CMakeLists.txt|*.cmake)(.N))
       source_files=(${source_files:#(build_*)/*})
+      source_files=(${source_files:#(vendor)/*})
```
royshil, line 73: i don't think we should format other people's code...

src/update-checker/github-utils.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include <string>
+
+#include <plugin-support.h>
+#include <obs.h>
+
+#include "github-utils.h"
+
+static const char GITHUB_LATEST_RELEASE_URL[] =
+	"https://api.github.com/repos/royshil/obs-backgroundremoval/releases/latest";
+static const char USER_AGENT[] = "obs-backgroundremoval/1.0.3";
```
royshil: this must come from the existingplugin build...

src/update-checker/github-utils.cpp
```diff
@@ -0,0 +1,38 @@
...
+		std::string str;
+		CURLcode code;
+		curl_easy_setopt(curl, CURLOPT_URL, GITHUB_LATEST_RELEASE_URL);
+		curl_easy_setopt(curl, CURLOPT_USERAGENT, USER_AGENT);
+		curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION,
+				 writeFunctionStdString);
+		curl_easy_setopt(curl, CURLOPT_WRITEDATA, &str);
+		code = curl_easy_perform(curl);
+		curl_easy_cleanup(curl);
+		obs_log(LOG_INFO, "%s\n", str.c_str());
```
royshil: i think `blog` should be availale

royshil (changes requested):
excellent stuff!!
but let's talk through it a bit

CMakeLists.txt
umireon: Originally I want ONNX Runtime and OpenCV to be submodules because it is easy to debug.
However, their library definitions are odd and it is hard to use add_subdirectory.
Therefore I have concluded that they should reside in ExternalProject.

On the other hand. the library definition of libcurl is straight-forward and it can be added by target_link_libraries.
I want to stick to the simpler way this time..

build-aux/run-cmake-format
umireon: This line instructs cmake-format to exclude the vendor directory.
This is not adding the vendor directory but removing vendor directory.

src/update-checker/github-utils.cpp
umireon: I couldn't find a way to concatenate the string stored in static variables in the compile time.
Do you know any good way?

royshil: you can just use a `.cpp.in` "config" file and use CMake's `config` to fill in the version from the CMake variables and generate the real `.cpp` file
but you can do the concatenation later e.g. with `std::string(USER_AGENT_PREFIX) + std::string(PLUGIN_VERSION)` or similar when you go to fetch the URL

src/update-checker/github-utils.cpp
umireon: Fixed

CMakeLists.txt
umireon: Fixed

src/update-checker/github-utils.cpp
umireon: Fixed

royshil (approved):
LGTM!

PR #366, opened 2023-07-02
Add depth estimation model

royshil:
Adding the TCMonoDepth model for depth estimation along with the "blur focus point" setting to simluate depth of field


umireon:
@royshil Please format

royshil:
@umireon this is ready for review

umireon (approved):
Amazing work!

royshil:
@umireon i've fixed the blur to be a "Focal blur", now it works as expected
Can you please check on your end?

umireon (changes requested):
Please fix the indentation in effect files.
Tabs and spaces are mixed but we used only tabs so far.

umireon:
And it does not work with the following error

```
error: effect_setval_inline: invalid param
```

royshil:
@umireon i've fixed the problems, thank you for pointing it out!

umireon (approved):
LGTM!

umireon:
We can call v1.1.0 for this feature?

PR #367, opened 2023-07-04
Remove QtNetwork

umireon:
We decided to use libcurl and not to use QtNetwork so I want to remove QtNetwork from this plugin.


royshil (approved):
go!

Issue #352, opened 2023-06-25
A way to notify users of a new plugin version


royshil:
**Is your feature request related to a problem? Please describe.**

I'd like plugin users to know when a new version drops and let them download it, or at least provide a download link

**Describe the solution you'd like**

A popup when the plugin loads and checks for new version that shows the new version download option.

**Describe alternatives you've considered**

I believe other plugins have a mechanism like that. I know OBS itself has it

**Additional context**

This can increase the success of users by using the latest versions

umireon:
I don't think OBS itself provides networking API but we can still use the networking C++ libraries such as Boost.ASIO.

royshil:
i'm not a fan of boost b/c it's a huge library, even if you take just a small part..
for smaller HTTP libs theres
* https://github.com/libcpr/cpr
* https://github.com/yhirose/cpp-httplib
* Qt - which is native in OBS already..

however i'm more concerned about how to pop -up a message for the user.. i don't know how that works

umireon:
We can use Qt directly.
https://github.com/obsproject/obs-plugintemplate/blob/master/CMakeLists.txt#L36

royshil:
We should have a dialog that says there's a new version and provides the user with a link to the downloads page on github.

umireon:
It would be great to show the release notes on the dialog as the OBS itself does

royshil:
IDK, it might be too much for the small dialog.. we would also want to cut it down to just the changelog
but it's certainly possible

Technetium1:
I hope this helps, should be able to get the changelog from the body if the releases are consistently labeled: https://api.github.com/repos/royshil/obs-backgroundremoval/releases/latest 

I use it to get the latest releases for my updater (python): https://github.com/Technetium1/ChocolateyUpdate/blob/2e3dbafb9b54bb4520505b9fa5488e9e5e016071/ChocolateyUpdate.py#L108

royshil:
we're pretty consistent on using `What's Changed` since it's auto generated by github in the release comments
so we can parse that out of the `body` in the response JSON

umireon:
We can use scrollable textarea for changelog as OBS

royshil:
i'm going to keep this issue open so we can iterate on the update dialog format
but this shouldn't keep us from shipping 1.1.0

PR #370, opened 2023-07-05
Update version check and dialog

royshil:
Implementing the version check and update dialog including the "opt out" option for update checks

towards #352


umireon (approved):
LGTM

PR #371, opened 2023-07-06
Fix releasing for macOS

umireon:
This is an upstream issue.
https://github.com/obsproject/obs-plugintemplate/pull/83


royshil (approved):
perfect. thanks!

PR #372, opened 2023-07-06
Add segment tracing AWS XRay

royshil:
Adding segment tracing with AWS Xray, including a dialog and code instrumentation


royshil:
i think we should probably release 1.1.0 without this addition
since it will be hard to pull together...

Issue #376, opened 2023-07-10
Update AUR packages


umireon:
We released v1 so we need to update AUR definition.

The current maintainer of AUR is @ngoonee so we have to contact them.

https://aur.archlinux.org/packages/obs-backgroundremoval
https://aur.archlinux.org/packages/obs-backgroundremoval-git

ngoonee:
I'm on it.

umireon:
What is 12.2? And please please post the log here to let me onvestigate the problem

ngoonee:
> What is 12.2? And please please post the log here to let me investigate the problem

Sorry, did not mention, its the Cuda version

CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find CUDA: Found unsuitable version "12.2", but required is exact
  version "12.1" (found /opt/cuda)
Call Stack (most recent call first):
  /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:598 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake/Modules/FindCUDA.cmake:1268 (find_package_handle_standard_args)
  /usr/lib/cmake/opencv4/OpenCVConfig.cmake:86 (find_package)
  /usr/lib/cmake/opencv4/OpenCVConfig.cmake:108 (find_host_package)
  CMakeLists.txt:60 (find_package)

umireon:
It seems that building of ONNX Runtime failed.
Can you make PKGBUILD depend on AUR ONNX Runtime?
You can use -DUSE_SYSTEM_ONNXRUNTIME=ON to load the system ONNX Runtime.

umireon:
I tried to reproduce this problem and hit another problem. This problem is related to libcurl and would be fixed in #377.

umireon:
@ngoonee Can you update obs-backgroundremoval-git as follows?

```
# Maintainer: Ng Oon-Ee <n g o o n e e dot t a l k @ gmail dot com>
# Based on obs-backgroundremoval PKGBUILD by PedroHLC <root@pedrohlc.com>
pkgname=obs-backgroundremoval-git
pkgver=1.1.0.r2.aeab837
pkgrel=1
pkgdesc="Background removal plugin for OBS studio"
arch=(x86_64)
url='https://github.com/royshil/obs-backgroundremoval'
license=('GPL2')
depends=('obs-studio' 'opencv' 'curl')
[...]

PR #377, opened 2023-07-10
Add the USE_SYSTEM_CURL flag

umireon:
Related #376
Using the system curl is an easier way to build our plugin on Arch Linux.


royshil (approved):
awesome stuff

PR #381, opened 2023-07-11
Fix so installation on Debian

umireon:
This is an upstream bug.
https://github.com/obsproject/obs-plugintemplate/pull/84


royshil (approved):
cool. thanks!

PR #384, opened 2023-07-12
Fix build-* and package-* to be symlink

umireon:
The build-* and package-* are symlinks on obs-plugintemplate but they were the actual files mistakenly.


royshil (approved):
Looks good

PR #385, opened 2023-07-12
Install obs-studio from ppa

umireon:
We need to take the upstream change.
https://github.com/obsproject/obs-plugintemplate/commit/13c6bbbaafcfef9547fdf926a1389f519126be70


royshil (approved):
Looks good

Issue #389, opened 2023-07-15
Provide the zip for a portable OBS installation on Windows


iNgeon:
If i my ask why is there no version of the windows zip file.
Some of us use OBS portable and would like to still extract it into those versions.

https://obsproject.com/forum/resources/obs-and-obs-studio-portable-mode-on-windows.359/

https://portableapps.com/apps/music_video/obs-studio-portable

umireon:
@iNgeon Here is the zip archive of 1.1.2.
https://github.com/royshil/obs-backgroundremoval/suites/14319111345/artifacts/805678926

I have just created #390 to include a zip archive on releases.
https://github.com/royshil/obs-backgroundremoval/pull/390
Starting from 1.1.3, a release will include a zip archive for portable installations of OBS on Windows.

iNgeon:
Thank you

PR #390, opened 2023-07-15
Build Zip

umireon:
As the request in #389

### DIFF
--- a/.github/workflows/build-project.yaml
+++ b/.github/workflows/build-project.yaml
@@ -196,13 +196,6 @@ jobs:
           target: x86_64
           config: ${{ needs.check-event.outputs.config }}
 
-      - name: Package Zip 📀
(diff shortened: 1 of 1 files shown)


.github/workflows/build-project.yaml
```diff
@@ -256,6 +256,13 @@ jobs:
           config: ${{ needs.check-event.outputs.config }}
           package: ${{ fromJSON(needs.check-event.outputs.package) }}
 
+      - name: Zip Plugin 📀
+        uses: ./.github/actions/package-plugin
+        with:
+          target: x64
```
royshil, line 262: But which OS is this for?

royshil (approved):
I think it's OK
But which OS is it for?
Only Linux? 
What about Windows?

umireon:
Im intended it only for Windows because an Windows user requested it.

PR #395, opened 2023-07-21
Fix links in build-aux

umireon:
The files in build-aux should be links to .run-format.zsh and this PR fixes that.


royshil (approved):
looks good!

Issue #393, opened 2023-07-18
Compile plugin on Fedora 37


MagnoBooter:
**Describe the bug**

Compilation errors on Fedora 37

**To Reproduce**

1. Clone source
2. Run `.github/scripts/build-linux --skip-deps` on Fedora 37

**Expected behavior**

The plugin should compile without errors.

**Compile Log**

Last messages before compilation stops:
```
[...]
[7/8] Building CXX object CMakeFiles/obs-backgroundremoval.dir/src/ort-utils/ort-session-utils.cpp.o
FAILED: CMakeFiles/obs-backgroundremoval.dir/src/ort-utils/ort-session-utils.cpp.o 
[...]

PR #399, opened 2023-07-21
Use system curl on Linux by default

umireon:
Related to #393


royshil (approved):
i think this is reasonable. although i'm not worried too much about building libcurl
it will make building faster though

PR #413, opened 2023-07-31
Embed the tutorial video on README.md

umireon:


royshil (approved):
in general i'm OK with putting the video embedded, however I think I will create a new one soon that's more complete and shows recent features...

PR #420, opened 2023-08-11
Remove libcurl from Mac and Windows

umireon:
We don't need to bundle libcurl on Mac and Windows because these OSes have their native HTTP clients.

The signature must include a callback because QTimer must be executed on the main thread.


umireon:
I confirmed this works on Windows and Mac

umireon:
I confirmed it works on Linux too.

royshil:
i actually think this is not necessary
first - libcurl seems to work perfectly fine, and it's well supported cross-platform
second - the savings in build time are not that great, about 2 minutes. a much bigger savings is from the prebuilt static OpenCV libs
and also - using libcurl made us have a unified simple API for fetching data which is easier to maintain than platform-specific code.

let me know what you think..

umireon:
I agree with your thought.

umireon:
@royshil This change saves 6 minutes on Mac and this saves time more than pre-built OpenCV.
https://github.com/royshil/obs-backgroundremoval/actions/runs/5838854087
https://github.com/royshil/obs-backgroundremoval/actions/runs/5838815434

royshil (approved):
OK we can roll with this

PR #430, opened 2023-08-27
C17 is only supported CMake 3.21+

umireon:
I have a build issue on openSUSE Leap 15.5 that its build system does not support C17.
The CMake provided by openSUSE Leap 15.5 is 3.20 and this version does not understand C17.
https://cmake.org/cmake/help/latest/prop_tgt/C_STANDARD.html
This PR will fix the build on openSUSE Leap.


royshil (approved):
Looking good

PR #445, opened 2023-09-22
Remove USE_SYSTEM_CURL

umireon:
The cache variable USE_SYSTEM_CURL is no longer used.
This PR removes this cache variable from CMakePresets.json


royshil (approved):
lgtm

PR #447, opened 2023-09-22
Bump onnxruntime to 1.16.0

### LINKED ISSUE #442: Upgrade to new ONNX Runtime 1.16
ONN Runtime 1.16 was just released
We can upgrade for performance and stability improvements.
https://github.com/microsoft/onnxruntime/releases/tag/v1.16.0

umireon:
Closes #442


src/onnxruntime_float16.h
```diff
@@ -0,0 +1,596 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
royshil: lets put thi file under `src/3rdparty` or `vendor` since it was not written by us

additionally - can we just fetch this file online during cmake instead of having it hardcoded here?

royshil (approved):
in general this is cool - good job!
but about the extra `.h` file see my comment

src/onnxruntime_float16.h
umireon: Downloading header files makes building on FlatPak and openSUSE so hard and then I don't want to fetch header files on build time.

PR #453, opened 2023-09-29
Remove the Packages.app

umireon:
We no longer use the Packages.app to create pkg on Mac.
The hash of Packages.app is changed and it causes the build failure so that I want to remove the Packages.app.

### DIFF
--- a/buildspec.json
+++ b/buildspec.json
@@ -32,12 +32,6 @@
         }
     },
     "tools": {
-        "packages": {
-            "version": "1.2.10",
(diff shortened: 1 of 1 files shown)


royshil (commented):
@umireon this is being used though in our build scripts, i believe, no?

umireon:
Packages.app is not used on our build script. I replaced the usage of the Packages.app with the commands that is included in Xcode such as pkgbuild and productbuild.

royshil (approved):
OK - Go!

PR #460, opened 2023-10-04
Add a user-friendly download page like OBS

umireon:
I have often seen Windows users are confused with so many download files on the release page and they mistakenly use zip archive.
Other Linux users are confused with the supported OBS version. Many mistakenly install an older OBS and complain on GitHub issues that our plugin is not working.
I suppose that the download page of our plugin needs to remind people of the prerequisites on the README.md prominently to reduce the issues related to the dependency and the installation process.

This PR will add the SvelteKit site that can be hosted on GitHub Pages.
The deploy preview is available on my fork.

https://umireon.github.io/obs-backgroundremoval/

The contents should be blushed up after I merge this PR. I intend the link to the latest release on README.md should be replaced with this page.
You have to turn on the GitHub Pages feature with the option of using GitHub Actions to deploy pages on the repository setting.
The URL of the site will be the following.

https://obs-ai.github.io/obs-backgroundremoval/


royshil (approved):
this is great!

we shuold definitely work on the content and design, but this is a good starting point

PR #461, opened 2023-10-05
Replace the download link with the site

umireon:
The link on the OBS forum would also be great to be replaced with this link.


royshil (approved):
checks out

Issue #454, opened 2023-09-30
Core dump on OBS startup with v1.1.6


Drizzt321:
**Describe the bug**

OBS-Studio core dumps during startup

**To Reproduce**
Debian 12.1, all up to date, KDE Plasma.

I did have it work a few weeks ago when I first installed v1.1.5, don't think any major libraries have been updated/installed, but I could be wrong.

Startup OBS from the command line. I had 1.1.5 installed, uninstalled that and installed the latest `.deb`, still same thing. 

```
$ obs --verbose
QSocketNotifier: Can only be used with threads started with QThread
debug: Found portal inhibitor
debug: Attempted path: share/obs/obs-studio/locale/en-US.ini
debug: Attempted path: /usr/share/obs/obs-studio/locale/en-US.ini
debug: Attempted path: share/obs/obs-studio/locale.ini
debug: Attempted path: /usr/share/obs/obs-studio/locale.ini
debug: Attempted path: share/obs/obs-studio/themes/Yami.qss
debug: Attempted path: /usr/share/obs/obs-studio/themes/Yami.qss
info: Command Line Arguments: --verbose
info: Platform: Wayland
info: CPU Name: AMD Ryzen 7 5800X3D 8-Core Processor
info: CPU Speed: 3549.195MHz
info: Physical Cores: 8, Logical Cores: 16
info: Physical Memory: 96497MB Total, 4553MB Free
info: Kernel Version: Linux 6.1.0-12-amd64
info: Distribution: "Debian GNU/Linux" "12"
info: Session Type: wayland
info: Qt Version: 5.15.8 (runtime), 5.15.8 (compiled)
info: Portable mode: false
debug: Attempted path: share/obs/obs-studio/themes/Dark/no_sources.svg
debug: Attempted path: /usr/share/obs/obs-studio/themes/Dark/no_sources.svg
QMetaObject::connectSlotsByName: No matching signal for on_tbar_position_valueChanged(int)
info: OBS 29.0.2.1-1+b1 (linux)
info: ---------------------------------
info: ---------------------------------
info: audio settings reset:
        samples per sec: 48000
        speakers:        2
        max buffering:   960 milliseconds
        buffering type:  dynamically increasing
info: ---------------------------------
info: Initializing OpenGL...
info: Using EGL/Wayland
info: Initialized EGL 1.5
[...]

PR #469, opened 2023-10-09
Fix updater crash

royshil:


royshil:
@umireon  are you able to review this? the folks in #454 have confirmed it resolves their bug

umireon (approved):
LGTM

PR #473, opened 2023-10-14
Fix wrong dependency name

saif-ellafi:
ffmpeg-6-libavcodec-devel fixed for OpenSUSE


saif-ellafi:
Fully tested and fixed in https://github.com/obs-ai/obs-backgroundremoval/pull/474 instead

PR #483, opened 2023-10-31
Update README.md

royshil:
we need to change all refernces to "OBS", i was asked by the OBS team
they are worried about conflicts


umireon (approved):
LGTM

PR #498, opened 2023-12-16
Add simple temporal smoothing

royshil:
To handle some of the flickering that often happens with some models


royshil:
@umireon What do you think?

umireon:
I will try this change this week

royshil:
> I will try this change this week

@umireon perfect! 
i was hoping to release v1.1.8 very soon

umireon:
I understand that this basically implements a low-pass filter for matt.

royshil:
> I understand that this implements a low-pass filter for matt.

Sort of, yes, but temporal. Really all it does is linearly blend the last mask with the current one.
But since the last mask is also a blend it's an "infinite impulse response".

umireon:
I'm doing z-transformation. Please wait.

umireon:
I think the inverse of the current smooth factor value corresponds to the cut-off frequency and thus I suppose the value of the factor of addWeighted should be inverted on the configuration screen.

royshil:
> I think the inverse of the current smooth factor value corresponds to the cut-off frequency and thus I suppose the value of the factor of addWeighted should be inverted on the configuration screen and it will work more straightforwardly.

the way i have it is that 1.0 corresponds to "no temporal filter" everything is taken from the current mask
and 0.0 is the exact opposite.
does that make sense?

umireon:
> the way i have it is that 1.0 corresponds to "no temporal filter" everything is taken from the current mask
and 0.0 is the exact opposite.
does that make sense?

It makes sense. Thank you!
What I want to say in the previous comment is actually that the effect of the smooth factor on the frequency domain is exponential and users may want to configure the factor with the logarithmic scale because it will have an intuitive behavior.

In addition, the addWeighted operation can be done on GPU shader and it would be more efficient than OpenCV's function.

royshil:
i think this cv operation on the CPU is very quick
the problem with moving it to the GPU/shader is that i need to keep/save the blended image for the next frame in CPU memory... so it's not easily done (i could move the rendered blended mask... but that saves very little in time)

royshil:
@umireon  can you approve so i can merge it in?

umireon (approved):
LGTM

Issue #499, opened 2023-12-20
Install triggers smart screen!


MrHinsh:
I have added your app to Winget:

https://github.com/microsoft/winget-pkgs/pull/131370

I'll submit a PR with an auto updater!

MrHinsh:
Also on manual download:

![image](https://github.com/occ-ai/obs-backgroundremoval/assets/5205575/8bc204d2-1b83-408b-8680-a83c9a19095e)

umireon:
SmartScreen will disappear when many people download our plugin successfully. We recently changed the distribution URL and Microsoft has not yet trusted our new page. I suppose we can do nothing about SmartScreen.

MrHinsh:
![image](https://github.com/occ-ai/obs-backgroundremoval/assets/5205575/139b6df5-92e9-48a5-a6c9-fbf04f0ffad4)

We will just have to wait!

MrHinsh:
I belive its likley being triggered becasue the "Comapny Name" is missing from the meta data.

You can also submit your application to https://learn.microsoft.com/en-us/windows/security/operating-system-security/virus-and-threat-protection/microsoft-defender-smartscreen/#submit-files-to-microsoft-defender-smartscreen-for-review

umireon:
Oh thanks! Can you make a PR to add company name to the metadata?

MrHinsh:
Sorry, I have no idea how to do that with CMake, which I have never used.

royshil:
I think this needs to be fixed in the OBS plugin template files. 
I can try to dig around. But so far my research came up short, it looks like the installer is in fact filling up the author information.

@MrHinsh How do you find those details?
If I know how I can make experiments and see if it affects the output

MrHinsh:
It had author, but no company. 

I just googled "view meta data in executable files on windows" and used a tool. I'm not at my computer to check, but it was installable with Winget.

I compared it to my file that downloads fine without triggering SmartScreen.

Did you submit your app to the smart screen check site?

MrHinsh:
:) Back at my computer! 

I used "exiftool".

```
❯ exiftool "C:\Users\MartinHinshelwood(He\Downloads\obs-backgroundremoval-1.1.8-windows-x64-Installer.exe"
ExifTool Version Number         : 12.70
File Name                       : obs-backgroundremoval-1.1.8-windows-x64-Installer.exe
Directory                       : C:/Users/MartinHinshelwood(He/Downloads
[...]

PR #505, opened 2023-12-28
Add metadata to suppress SmartScreen on Windows

umireon:
Related to #499


royshil (approved):
For the copyright can we use a variable for app author?
If so - we can also submit this to the OBS plugin template

umireon:
@royshil The copyright text includes the year. What should I do about year?

royshil:
@umireon I think it's ok
Let's start with this..
We could have cmake plug in the year. It shouldn't be difficult

PR #515, opened 2024-01-06
Fix Flatpak build

umireon:
Onnxruntime_VERSION is needed when custom tarball is used


umireon:
I need 1.1.10 to update Flatpak after this PR is merged.

royshil (approved):
lgtm!

PR #528, opened 2024-01-13
Improve the handling of path string

### LINKED ISSUE #525: suspected use of bad pointer in log message
When running obs from commanline (ubuntu 22.04, obs 30.0 ), I saw this line:
```
info: [obs-backgroundremoval]   Model file path: H�	S
```

It looks like using uninit pointer.

**To Reproduce**
install the plugin ( I used sudo dpkg -i obs-backgroundremoval-1.1.10-x86_64-linux-gnu.deb )

Run OBS from cli.
Close the program (from GUI).
read the terminal text

**Expected behavior**
show empty string.

umireon:
Closes #525 

Given `std::string s`, `s.c_str()` will be released when `s` is released. If you want to use `s.c_str()` later, you have to hold `s` until you use `s.c_str()`.

On Windows, char paths are stored as multi-byte characters which Windows defines and it is hard to manipulate them with C++ standard libraries (codecvt was deprecated in C++17) so I propose to convert multi-byte characters with Windows-specific API that is guaranteed to work with multi-byte characters on Windows.

### DIFF
--- a/src/FilterData.h
+++ b/src/FilterData.h
@@ -30,9 +30,9 @@ struct filter_data : public ORTModelData {
 	std::mutex outputLock;
 
 #if _WIN32
-	const wchar_t *modelFilepath = nullptr;
+	std::wstring modelFilepath;
 #else
-	const char *modelFilepath = nullptr;
+	std::string modelFilepath;
 #endif
 };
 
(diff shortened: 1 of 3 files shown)


royshil (dismissed):
This checks out
But make sure you test on Windows with a model path that has Unicode characters in it . This caused a bunch of problems in the past

umireon:
I will check on Windows but it uses std::wstring on Windows.

umireon:
I have checked it on Windows having multi-byte characters on the path and noticed that the current path handling wouldn't work under some conditions so I have added a code to handle multi-byte characters properly.

Please review again, @royshil

royshil (approved):
looks good!

PR #558, opened 2024-03-14
Fix image similarity calculation in background filter

royshil:


umireon (approved):
LGTM

umireon:
Please do not remove or retag 1.1.11 but create 1.1.12

PR #565, opened 2024-03-19
Update Onnxruntime version and fix Windows compilerconfig

royshil:


royshil:
@umireon we have a problem with onnxruntime 1.17.1 on linux where the .tgz file doesn't have the `include/tensorrt_provider_factory.h`
in v1.16.3 it did exist
i wonder if this file is no longer needed but TensorRT support still remains. it seems so because the .so file for tensorrt EP indeed exists in v1.17.1

umireon:
@royshil `SessionOptionsAppendExecutionProvider_TensorRT` and `SessionOptionsAppendExecutionProvider_TensorRT_V2` got to be included in onnxruntime_c_api.h so that we can safely remove the inclusion of `include/tensorrt_provider_factory.h` without any changes.

royshil:
@umireon the build is good - do you want to test or approve? thanks!

umireon (approved):

PR #575, opened 2024-04-23
Add CUDA execution provider and enable TensorRT engine cache

### LINKED ISSUE #549: Cache TensorRT modules
**Is your feature request related to a problem? Please describe.**

Every time I open the OBS with this plugin installed, there's a 60 sec+ delay while the TensorRT modules are compiled.

**Describe the solution you'd like**

Most applications use a cache to prevent this from occurring on each startup. That would be great.

**Describe alternatives you've considered**

Waiting a long time each time I start OBS, which is annoying.

krakowski:
## Description

This PR adds the following new functionality for Linux based systems:

  * In addition to the TensorRT Execution Provider, the CUDA Execution Provider can now also be selected via the GUI. The loading process is significantly faster than loading the TensorRT Engine (for the first time), as the model does not have to be optimized beforehand.

  * The TensorRT Execution Provider now uses a cache so that the model does not have to be recreated every time. This reduces the start time from ~ 50 seconds to ~ 4 seconds (in my case).

I have also made a small change to the build script, allowing users to set the ONNX Runtime version using variables.

## Related Issues

Fixes #549


umireon:
Please remove the CUSTOM_ONNXRUNTIME_VERSION variable. We don't want to support arbitrary combinations of our plugin and ONNX Runtime.

umireon:
Is the change of cmake/FetchOnnxruntime.cmake needed for the CUDA provider and cache for the TensorRT provider.
If this change is not essential, please remove the changes from this PR and include it in another PR.

krakowski:
> Please remove the CUSTOM_ONNXRUNTIME_VERSION variable. We don't want to support arbitrary combinations of our plugin and ONNX Runtime.

Sure, I will remove the commit and force push to my branch :+1: 

Some background why I added this:

ONNX Runtime 1.17.1 is compiled against CUDA 11, while my system has CUDA 12 installed. By changing the URL (https://github.com/microsoft/onnxruntime/releases/download/v1.17.3/onnxruntime-linux-x64-gpu-cuda12-1.17.3.tgz) and version (`1.17.3`), I was able to use the plugin with CUDA 12. The archive's naming convention doesn't seem to be stable, so I couldn't find a way around specifying everything manually.

  * `1.17.3` : https://github.com/microsoft/onnxruntime/releases/download/v1.17.3/onnxruntime-linux-x64-gpu-cuda12-1.17.3.tgz
  * `1.17.1` : https://github.com/microsoft/onnxruntime/releases/download/v1.17.1/onnxruntime-linux-x64-cuda12-1.17.1.tgz

umireon:
Since we do not require the contributions to be contained in a single commit, please do not force-push for later reference.

umireon:
I suppose the build script will accept the ONNXRUNTIME_PROVIDER variable and choose the corresponding predefined URL rather than accept any ONNX Runtime version.

krakowski:
I placed the deleted commit in a separate branch for later reference: https://github.com/occ-ai/obs-backgroundremoval/compare/main...krakowski:obs-backgroundremoval:refactor/build

umireon:
I think cpu, cuda11, and cuda12 providers are available on the official distributions.

umireon:
Please run ./build-aux/run-clang-format and ./build-aux/run-cmake-format to fix the formatting errors.

umireon:
I think we first need to update our ONNX Runtime version to be 1.17.3 to merge this PR, right?

krakowski:
> I think we first need to update our ONNX Runtime version to be 1.17.3 to merge this PR, right?

This should work with ONNX 1.17.1, too. I changed the version (which is now reverted) because version 1.17.1's CUDA 12 archive (https://github.com/microsoft/onnxruntime/releases/download/v1.17.1/onnxruntime-linux-x64-cuda12-1.17.1.tgz) is missing a shared library.

image

umireon:
Sure I will update our ONNX Runtime to 1.17.3 after this PR is merged.

src/ort-utils/ort-session-utils.cpp
```diff
@@ -24,6 +25,7 @@ int createOrtSession(filter_data *tf)
 		return OBS_BGREMOVAL_ORT_SESSION_ERROR_INVALID_MODEL;
 	}
 
+	const auto &api = Ort::GetApi();
```
umireon:
```suggestion
```

src/ort-utils/ort-session-utils.cpp
```diff
@@ -64,8 +66,51 @@ int createOrtSession(filter_data *tf)
 #if defined(__linux__) && defined(__x86_64__) && \
 	!defined(DISABLE_ONNXRUNTIME_GPU)
 		if (tf->useGPU == USEGPU_TENSORRT) {
+
```
umireon, line 69:
```suggestion
			const auto &api = Ort::GetApi();

```

umireon (changes requested):

umireon:
This suggestion will fix the build problem on Mac.

src/ort-utils/ort-session-utils.cpp
```diff
@@ -64,8 +65,52 @@ int createOrtSession(filter_data *tf)
...
+			// Initialize TensorRT provider options
+			OrtTensorRTProviderOptionsV2 *tensorrt_options;
+			Ort::ThrowOnError(api.CreateTensorRTProviderOptions(
+				&tensorrt_options));
+
+			// Create cache folder if it does not exist
+			std::filesystem::path cache_folder(tensorrt_cache_path);
+			if (!std::filesystem::exists(cache_folder)) {
+				std::filesystem::create_directories(
```
royshil, line 82: what if this fails?

royshil:
@krakowski thanks for the great work

i'm not asking to do anythign further in this PR
by if we're here aleady we may alerady add CUDA support for windows... i mean its just packaging the provider .dll for onnxrt

src/ort-utils/ort-session-utils.cpp
krakowski: If this fails the execution provider will throw an error on initialization (file not found).

In my opinion this shouldn't fail, since it uses a home directory location (`~/.cache`), which should be accessible by the current logged in user. I took the code from here: 

https://github.com/occ-ai/obs-backgroundremoval/blob/main/src/obs-utils/obs-config-utils.cpp#L18-L29

umireon (approved):

PR #594, opened 2024-08-03
Fix memory leaks

itmo153277:


itmo153277:
This code seem to be copy-pasted to the other obs plugins by locaal-ai, the fix should be applied to them as well, I am too lazy to do it though.

royshil (approved):
awesome! thanks!

PR #597, opened 2024-08-13
fix create_config_folder not compatible with non-ansi paths

epessoa-i:
Thanks for your work on this module.
I noticed the module makes obs crash when the config folder has non-ansi chars on it.
So, I'm proposing this small change, that makes the folder creation similar to modules that run successfully.


royshil (approved):
this checks out!

Issue #591, opened 2024-07-16
[LINUX][NVIDIA] Doesn't work properly with NVIDIA GPU 


mercifulboss:
When using the "CPU" option, the background remover updates as you move around and continues removing the background. When you switch to "GPU" option, it updates the silhouette once and then doesn't update or do anything after the first time.

**To Reproduce**

1. Go to Background Removal Filter on OBS
2. Click on "GPU"
3. See it doesn't update the background

**Expected behavior**

Supposed to update the background using the "GPU" option

**Desktop (please complete the following information):**

OS: Arch Linux x86_64 
Kernel: 6.9.9-arch1-1 
DE: Plasma 6.1.2 
CPU: Intel i7-10750H (12) @ 5.000GHz
GPU: NVIDIA GeForce RTX 2060 Mobile
Browser: Firefox
OBS Version: 30.2.0

royshil:
can you please share the OBS logs so we can investigate?

nessvan:
Thank you Roy for the great work!
I have the same issue:
![image](https://github.com/user-attachments/assets/94e418c9-923b-42bf-ae79-1387b1975c19)
this part of the log is for cpu where it works:
info: [obs-backgroundremoval] get_latest_version: 
info: [obs-backgroundremoval] Background filter updated
info: [obs-backgroundremoval] Model models/mediapipe.onnx input 0: name input_1:0 shape (4 dim) 1 x 144 x 256 x 3
info: [obs-backgroundremoval] Model models/mediapipe.onnx output 0: name segment:0 shape (4 dim) 1 x 144 x 256 x 2
info: [obs-backgroundremoval] Allocated 110592 sized float-array for input 0
info: [obs-backgroundremoval] Allocated 73728 sized float-array for output 0
info: [obs-backgroundremoval] Background Removal Filter Options:
info: [obs-backgroundremoval]   Source: Background Removal
info: [obs-backgroundremoval]   Model: models/mediapipe.onnx
info: [obs-backgroundremoval]   Inference Device: cpu
info: [obs-backgroundremoval]   Num Threads: 1
info: [obs-backgroundremoval]   Enable Threshold: true
info: [obs-backgroundremoval]   Threshold: 0.500000
info: [obs-backgroundremoval]   Contour Filter: 0.050000
info: [obs-backgroundremoval]   Smooth Contour: 0.500000
info: [obs-backgroundremoval]   Feather: 0.000000
[...]

PR #621, opened 2025-03-14
fix custom onnxruntime version

fritz-fritz:
Introduced a new variable so that built libraries appropriately reflect the custom onnxruntime version.
Also added a symlink from libonxruntime.so.1

The real benefit of this is seen when leveraging CMakeUserPresets as shown in [my gist here](https://gist.github.com/fritz-fritz/b8dd3cc081325d1a73e86be727edc0f7) that addresses #591.

This pull request does not update the default onnxruntime versions or integrate with the installer scripts.

### DIFF
--- a/cmake/FetchOnnxruntime.cmake
+++ b/cmake/FetchOnnxruntime.cmake
@@ -8,6 +8,10 @@ set(CUSTOM_ONNXRUNTIME_HASH
     ""
     CACHE STRING "Hash of a downloaded ONNX Runtime tarball")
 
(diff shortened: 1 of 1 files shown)


royshil:
@fritz-fritz can you run `cmake-format` on `cmake/FetchOnnxruntime.cmake` plz?

fritz-fritz:
@royshil Should be good to go now :heavy_check_mark:

royshil (approved):
looks good!

PR #656, opened 2025-11-03
Disabling codesign for Mac

umireon:
Configuring code signing for Mac have some problems so that we skip code signing for this release.


umireon:
No go for this PR because the problem was fixed.

PR #658, opened 2025-11-04
Refactor update-checker and use libcurl from vcpkg on all the platform

umireon:
This is a prior step to integrate this plugin with WinML (Windows-embedded ONNX Runtime).

The original implementation of WinRTHttpClient is based on very old technologies (C++/CX) and interferes with the implementation for WinML (C++/WinRT).

This PR removes the dependencies for this old technology and easier and more integrated access to the internet provided by only one library named libcurl within the plugin.

wolfSSL is the best SSL solution when developing GPL products and we are on GPLv3.

The modified wolfSSL port is included in this PR because the official wolfSSL port is broken when that is combined with libcurl. I've fixed the port and copied it from my plugin.


umireon:
This PR is for removing the dependency to C++/CX. Building has succeeded and I don't care the functionality of update checking is regressed. We can fix it later before the next stable release if it gets broken by this change.

umireon:
Added .github/scripts/install-vcpkg-macos.bash to build universal version of dependent libraries on Mac. This script is well-tested on my plugins and any libraries.

PR #659, opened 2025-11-05
Add credentials for vcpkg binary caches

umireon:
Provide GitHub credential to NuGet to retrieve vcpkg binary caches from vcpkg-caches


umireon:
Checking if the build caches works as expected.

umireon:
Caches work as expected.

PR #661, opened 2025-11-05
Install OpenCV by vcpkg

umireon:
Now OpenCV is managed by vcpkg.
pkg-config is required for Arch build.


umireon:
I cannot bypass review

umireon:
Confirmed to work on Mac.
Don't care on Windows because it is already broken before this PR.

PR #662, opened 2025-11-05
Cleaner ONNX Runtime downloader for Mac

umireon:
Improve Mac build

Don't care about Windows and Ubuntu now.


umireon:
@royshil Windows and Ubuntu are failing but they will be fixed later.

royshil (approved):
if you're OK with windows and ubuntu failig - GO!

PR #664, opened 2025-11-06
Fix Windows build with ONNX Runtime

umireon:
I have fixed Windows build and confirmed to work it stably.

DirectML support on ONNX Runtime is deprecated and having DirectML breaks our plugin so that DirectML support in our plugin is also removed.

Ubuntu build still broken and it will be fixed later.


umireon:
@royshil Managed to fix it on Windows.

royshil (approved):
Very cool!

PR #676, opened 2025-11-09
Refresh documents (README.md and our site)

umireon:
The documentation is updated to reflect current situation.
README.md was changed for users to find download links easily.
GPU support is not verified now so commented out.
Building instructions are outdated so removed.


royshil (approved):
Great work

PR #679, opened 2025-11-09
Bundle ONNX Runtime into deb package

umireon:
CMakeLists.txt was modified to include libonnxruntime.so into the deb package.


royshil (approved):
Very cool

PR #685, opened 2025-11-09
Add FAQ.md for better user support

umireon:
I'm considering about AI-based chat support app for this plugin to unburden us from the user's support request.
That will be WebLLM-based one and all the computation will be done in the user's computer locally.
I have generated knowledge base file for frequently asked questions by asking Copilot to collect them.
I have read all the parts of this file carefully and replaced false statement with other true statement as possible as I can.
This file itself would be useful for users who has a question about this plugin.
How do you think?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@
 !buildspec.json
 !CMakeLists.txt
 !CMakePresets.json
+!FAQ.md
 !LICENSE
 !README.md
 !vcpkg-configuration.json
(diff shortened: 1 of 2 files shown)


umireon:
Thinking as this one. No cost because everything will be done on user's computer.

<video src="https://github.com/user-attachments/assets/519adb36-45c0-455a-b743-3e2307a0019c" width="320"></video>

PR #688, opened 2025-11-10
Context for chatbot was compressed

umireon:
I have compressed the context of chatbot and choose the smaller model for faster talk.


umireon:
@royshil ping

PR #753, opened 2026-01-22
Add note about Lite version for background removal

umireon:
Add link to LBRL to clarify the relationship between our two plugins.
This will reduce the amount of issues we have to handle seriously.


royshil (approved):
great

PR #754, opened 2026-01-23
Remove AVX2-related options from vcpkg

umireon:
OpenCV has dynamic dispatch for AVX2 and we do not need to specify -mavx2 for compiler.
Removing -mavx2 allows users whose CPU don't have AVX2 options so that we can provide better compatibility.


royshil (approved):
alright!

PR #764, opened 2026-02-12
Migrate to the new install recommendation for OBS plugins

umireon:


umireon:
@royshil please review

PR #765, opened 2026-02-13
Static-linking ONNX Runtime on Mac to avoid duplicate symbols

umireon:
Add build scripts for reduced ONNX Runtime build for Mac.

- Build caches for ORT's vcpkg and ORT itself are enabled
- Flat ORT models are regenerated for reduced ONNX Runtime library
- Link reduced ORT and related dependencies on Mac
- Hide all the symbols other than OBS plugin requirements

Not to use Minimal Build or Extended Minimal Build because they are hard to integrate with CoreML.

### DIFF
--- a/.github/workflows/build-project.yaml
+++ b/.github/workflows/build-project.yaml
@@ -180,6 +180,7 @@ jobs:
 
       - name: Build ONNX Runtime
         run: |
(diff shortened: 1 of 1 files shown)


umireon:
@royshil Please review

PR #766, opened 2026-02-14
Switch to runtime-optimized .ort model files

umireon:
I'm considering to link a minimal reduced ONNX Runtime library statically to this plugin for better compatibility, less build duration, and for preventing this plugin from conflicting against other ORT-based plugins such as LocalVocal.

Converting ONNX format into ORT format is required for minimal build of ONNX Runtime, so that I want to do it first. Standard ONNX Runtime can also read ORT format and this change works with the current code.

There are Fixed and Runtime optimization mode and I chose the Runtime optimization mode because Runtime mode lets CUDA or CoreML to optimize, although Fixed optimization mode only improve the startup of ORT session on CPU slightly.

Models can be converted with the following commands:

```
python3.11 -m venv .venv
. .venv/bin/activate
pip3 install onnx onnxruntime==1.24.1
python3 -m onnxruntime.tools.convert_onnx_models_to_ort --output_dir data/models --optimization_style Runtime -- models
```


umireon:
@royshil Please review

PR #768, opened 2026-02-17
Static-linking ONNX Runtime on Windows CPU build

umireon:
Related to #765

ORT with CUDA cannot be compiled as static library due to its complex dependencies so I will investigate other means later.

Ccache will be enabled after this PR built and merged successfully.


umireon:
@royshil Please review

PR #769, opened 2026-02-17
Enable ccache for Windows

umireon:
For less build duration


umireon:
@royshil Please review

PR #771, opened 2026-02-19
Static-linking ONNX Runtime on Ubuntu CPU build

umireon:
As well as other platforms.


umireon:
@royshil Please review

PR #773, opened 2026-02-20
Remove source tarball packaging and upload

umireon:
Remove creating source.tar.xz because it includes binaries, and we keep our code buildable with git clone.


umireon:
@royshil Please review

royshil (approved):
Makes sense

PR #774, opened 2026-02-20
Add ccache variables for correct caching on Ubuntu

umireon:
Adding reasonable settings for Linux


umireon:
@royshil Please review

royshil (approved):
Ok!

PR #776, opened 2026-02-20
Improve vcpkg dependency installation

umireon:
- Use vcpkg in the runner image for better integrity
- vcpkg_installed is now under .deps_vendor for consistency with ONNX Runtime
- install-vcpkg-macos.bash is removed, and integrated with merge_vcpkg_installed_into_macos_universal.sh


umireon:
@royshil Please review

umireon:
@royshil ping

PR #780, opened 2026-02-20
Fix old repo names

umireon:
Replace occ-ai and locaal-ai with royshil.


umireon:
@royshil Please review

PR #784, opened 2026-02-26
workflow: attest artifacts and use gh release

umireon:
Improve release creation process

- Use GitHub Attestation for distribution genuineness
- Use gh command to reduce third-party dependency


umireon:
@royshil Please review

PR #786, opened 2026-02-26
Fix frame handoff race and eliminate redundant copies

xav-ie:
## Summary

- Replaces `inputBGRA.empty()` checks with an `std::atomic<bool> newFrameAvailable` flag, fixing a data race where the tick thread could read `inputBGRA` outside the mutex in `enhance-filter.cpp`
- Uses `cv::swap()` instead of `clone()` in both `background-filter.cpp` and `enhance-filter.cpp` for O(1) frame handoff from render thread to tick thread
- Uses `copyTo()` instead of `clone()` in `obs-utils.cpp` so the destination buffer is reused when dimensions haven't changed, avoiding a heap allocation per frame

## Benchmark results (300 iterations, best of 3 runs)

| Step | Median (µs/frame) | Change |
|------|-------------------|--------|
| Before (clone + empty() check) | 1127.7 | — |
| After (swap + copyTo + atomic flag) | 675.9 | −40% |

## Pull Request Checklist

- [ ] I have read the latest CONTRIBUTING.md.
- [ ] I have acknowledged the licensing and patent grant policy.
- [ ] I have signed off and verified all my commits.

### DIFF
--- a/src/background-filter.cpp
+++ b/src/background-filter.cpp
@@ -536,11 +536,11 @@ void background_filter_video_tick(void *data, float seconds)
 			// No data to process
 			return;
 		}
-		if (tf->inputBGRA.empty()) {
-			// No data to process
+		if (!tf->newFrameAvailable) {
 			return;
 		}
-		imageBGRA = tf->inputBGRA.clone();
+		cv::swap(imageBGRA, tf->inputBGRA);
+		tf->newFrameAvailable = false;
 	}
 
 	if (tf->enableImageSimilarity) {
--- a/src/enhance-filter.cpp
+++ b/src/enhance-filter.cpp
@@ -247,10 +247,6 @@ void enhance_filter_video_tick(void *data, float seconds)
 		return;
 	}
 
-	if (tf->inputBGRA.empty()) {
-		return;
-	}
-
 	// Get input image from source rendering pipeline
 	cv::Mat imageBGRA;
 	{
@@ -258,7 +254,11 @@ void enhance_filter_video_tick(void *data, float seconds)
 		if (!lock.owns_lock()) {
 			return;
 		}
-		imageBGRA = tf->inputBGRA.clone();
+		if (!tf->newFrameAvailable) {
+			return;
+		}
+		cv::swap(imageBGRA, tf->inputBGRA);
+		tf->newFrameAvailable = false;
 	}
 
 	cv::Mat outputImage;
--- a/src/FilterData.hpp
+++ b/src/FilterData.hpp
@@ -31,6 +31,7 @@ struct filter_data : public ORTModelData, public std::enable_shared_from_this<fi
 	gs_stagesurf_t *stagesurface;
 
 	cv::Mat inputBGRA;
+	std::atomic<bool> newFrameAvailable{false};
 
 	std::atomic<bool> isDisabled{false};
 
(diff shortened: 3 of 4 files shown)


royshil (commented):
I remember vaguely that we had used swap in the past and reversed back to clone... iirc the reason was that it ran into mem alloc issues

Time savings are great but we can't regress functionality or introduce bugs

xav-ie:
Thanks for the review! I searched the history and couldn't find evidence of `cv::swap` being used before:

```
$ git log origin/main -S "cv::swap" --oneline -- src/
(no results)
```

You might be thinking of PR #225, which proposed replacing `cv::Mat` with raw `uint8_t` buffers + `memcpy` for inter-thread communication — that was closed in favor of the mutex + `clone()` approach in PR #224.

Our `cv::swap` here is safe under the existing mutex:

- **Render thread** (`obs-utils.cpp`): holds `inputBGRALock`, calls `copyTo()` which writes into `tf->inputBGRA`'s buffer
- **Tick thread** (`background-filter.cpp`): holds `inputBGRALock`, calls `cv::swap(imageBGRA, tf->inputBGRA)` which just swaps the Mat headers (pointer, size, refcount) in O(1) — no allocation, no copy
- After the swap, the tick thread exclusively owns the pixel data and `tf->inputBGRA` is left empty
- ~~So the only allocation that ever happens is when frame dimensions change, which is the same as `clone()` but without re-allocating every single frame.~~

**Correction:** Since swap empties `tf->inputBGRA`, `copyTo` allocates fresh on the next render — so this halves allocations and copies (2 per cycle → 1), not eliminates them. That's where the ~40% benchmark improvement comes from.

umireon:
#224 and #225 are my PRs. Thanks for contributing stability!

umireon (dismissed):
Working on the CONTRIBUTING.md now. Sorry but wait a moment, please.

umireon:
@xav-ie I've added checklist to your description. Feel free to ask us for question, it is added to this project just now.

umireon:
Licensing works are almost completed. #797

umireon:
@xav-ie  #797 merged. Please update your branches and let me help you align with our guidelines.

royshil:
@xav-ie are you still working o n this as a draft or want to make it a legit pull requets?

src/background-filter.cpp
```diff
@@ -536,11 +536,11 @@ void background_filter_video_tick(void *data, float seconds)
 			// No data to process
 			return;
 		}
-		if (tf->inputBGRA.empty()) {
-			// No data to process
+		if (!tf->newFrameAvailable) {
```
umireon, line 539: The following code:

```
		if (!tf->newFrameAvailable) {
			return;
		}
		imageBGRA = tf->inputBGRA.clone();
		cv::swap(imageBGRA, tf->inputBGRA);
		tf->newFrameAvailable = false;
```

should be like:

```
		if (tf->newFrameAvailable.exchange(false, std::memory_order_acq_rel)) {
			cv::swap(imageBGRA, tf->inputBGRA);
		}
```

for better performance with appropriate memory barrier.

src/enhance-filter.cpp
```diff
@@ -247,18 +247,18 @@ void enhance_filter_video_tick(void *data, float seconds)
...
 		std::unique_lock<std::mutex> lock(tf->inputBGRALock, std::try_to_lock);
 		if (!lock.owns_lock()) {
 			return;
 		}
-		imageBGRA = tf->inputBGRA.clone();
+		if (!tf->newFrameAvailable) {
+			return;
+		}
+		cv::swap(imageBGRA, tf->inputBGRA);
+		tf->newFrameAvailable = false;
```
umireon, line 261:
```suggestion
		if (!tf->newFrameAvailable.exchange(false, std::memory_order_acq_rel)) {
			cv::swap(imageBGRA, tf->inputBGRA);
		}
```

PR #787, opened 2026-02-26
Optimize inference preprocessing pipeline

xav-ie:
## Summary

- Computes image similarity PSNR on a 192×108 thumbnail instead of the full frame, reducing the comparison cost by ~99% of pixels while preserving change detection fidelity
- Reorders the inference preprocessing to resize BGRA first and then convert to RGB, so `cvtColor` operates on the smaller network-input-sized image instead of the full frame
- Switches preprocessing resize interpolation from `INTER_LINEAR` to `INTER_NEAREST` since the result feeds directly into a neural network that is invariant to interpolation quality

## Benchmark results (300 iterations, best of 3 runs)

| Step | Median (µs/frame) | Change |
|------|-------------------|--------|
| Baseline | 2532.0 | — |
| + Thumbnail PSNR | 1731.4 | −32% |
| + Resize before cvtColor + INTER_NEAREST | 1127.7 | −35% |
| **Cumulative** | **1127.7** | **−55%** |

## Pull Request Checklist

- [ ] I have read the latest CONTRIBUTING.md.
- [ ] I have acknowledged the licensing and patent grant policy.
- [ ] I have signed off and verified all my commits.

### DIFF
--- a/src/ort-utils/ort-session-utils.cpp
+++ b/src/ort-utils/ort-session-utils.cpp
@@ -138,16 +138,16 @@ bool runFilterModelInference(filter_data *tf, const cv::Mat &imageBGRA, cv::Mat 
 		return false;
 	}
 
-	// To RGB
-	cv::Mat imageRGB;
-	cv::cvtColor(imageBGRA, imageRGB, cv::COLOR_BGRA2RGB);
-
-	// Resize to network input size
+	// Resize to network input size first (operate on smaller image)
 	uint32_t inputWidth, inputHeight;
 	tf->model->getNetworkInputSize(tf->inputDims, inputWidth, inputHeight);
 
+	cv::Mat resizedBGRA;
+	cv::resize(imageBGRA, resizedBGRA, cv::Size(inputWidth, inputHeight), 0, 0, cv::INTER_NEAREST);
+
+	// To RGB (on already-resized image)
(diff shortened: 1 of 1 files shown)


src/background-filter.cpp
```diff
@@ -544,16 +544,18 @@ void background_filter_video_tick(void *data, float seconds)
...
+		if (!tf->lastImageBGRA.empty() && tf->lastImageBGRA.size() == thumbnail.size()) {
+			double psnr = cv::PSNR(tf->lastImageBGRA, thumbnail);
 
 			if (psnr > tf->imageSimilarityThreshold) {
-				// The image is almost the same as the previous one. Skip processing.
 				return;
 			}
 		}
-		tf->lastImageBGRA = imageBGRA.clone();
+		tf->lastImageBGRA = thumbnail;
```
royshil: Is the last image only used for PSNR calculations?
If so I suppose it's fine

royshil (commented):
Did you test this with all models?

xav-ie:
> Did you test this with all models?

No, sorry. Let me try to figure this out. Sorry for the noise.

royshil:
@xav-ie do you want to bring this up so we can incorporate in the plugin? this is a very good contribution

PR #888, opened 2026-06-02
[codex] Improve mask edge softness post-processing

LeMegaGeek:
## Summary

- Extract background mask post-processing into a small reusable module.
- Add an `Edge softness` threshold setting so mask edges can keep partial alpha instead of becoming a hard binary contour.
- Resize the model mask to the source size before threshold/shape processing.
- Add a focused CTest target for mask post-processing behavior.

## Why

The current threshold path turns model output into a binary mask early, then smooths/re-thresholds it. That makes portrait edges look cut out, especially around hair, shoulders, and glasses. Keeping a soft transition around the threshold preserves more of the model matte while leaving a hard-threshold mode available by setting edge softness to zero.

## Validation

- `.venv/bin/clang-format --dry-run --Werror src/background-filter.cpp src/mask-post-processing.cpp src/mask-post-processing.hpp tests/mask-post-processing-test.cpp`
- `.venv/bin/gersemi --check CMakeLists.txt`
- `.venv/bin/reuse lint`
- `git diff --check`

Full local CMake configuration was not possible because `libobs` development files are not installed in this environment (`find_package(libobs)` fails before reaching this change).


LeMegaGeek:
Closing this draft for now while the contribution is reviewed locally.

PR #910, opened 2026-06-27
Fix false positive update alert when local version is newer

henriquegeremia:
### Description
This PR fixes a bug where the plugin incorrectly alerts the user that an update is available when the locally installed version is actually newer than the latest version fetched from the GitHub API. 

Currently, users running a newer local version (e.g., `1.4.0`) are prompted with `🚀 Update available! (1.3.7)` because the update checking logic in `update-checker.cpp` only checks for strict equality (`info.version == PLUGIN_VERSION`). If the versions do not match, it assumes a new update is available.

### Solution
1. Introduced a semantic version comparison helper `is_newer_version` that splits both the remote and local version strings by `.` and compares the integer components sequentially.
2. Updated the comparison inside the callback of `check_update` to only set `latestVersionForUpdate` if the fetched remote version is strictly newer than the currently installed `PLUGIN_VERSION`.

---

## Pull Request Checklist

- [x] I have included the required license headers.
- [x] I am confident with my code integrity.

### DIFF
--- a/src/update-checker/update-checker.cpp
+++ b/src/update-checker/update-checker.cpp
@@ -13,8 +13,7 @@
 #include "../plugin-support.h"
 
 #include <mutex>
-#include <sstream>
-#include <vector>
+#include <regex>
 #include <algorithm>
 
 extern "C" const char *PLUGIN_VERSION;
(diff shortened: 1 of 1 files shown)


umireon:
Nice catch!

umireon (commented):
You can use regex. This part is run only once and regex is preferred to stringstream.

henriquegeremia:
Done! I've updated the implementation to use std::regex instead of std::stringstream. Thanks for the feedback!

umireon (changes requested):
Please use the following regex:

```
^(\d+)\.(\d+)\.(\d+)(?:-([a-zA-Z]+)(?:\.?(\d+))?)?$
```

Simply ignore the 4th and 5th component.
And please make this PR ready.

umireon:
You are required to check the Pull Request checklist I've added. Do not remove template.

henriquegeremia:
Done! I've updated the regex to match the exact pattern you suggested, restored the PR checklist template, and marked the PR as ready for review.

umireon (approved):

umireon:
@henriquegeremia Thanks for contribution! Finally, please sign your commits and add DCO! See https://github.com/royshil/obs-backgroundremoval/blob/main/CONTRIBUTING.md#4-developer-certificate-of-origin-dco-and-commit-signing

umireon:
Stalled. Closing.

henriquegeremia:
@umireon Hi! I have signed the commit with SSH and added the DCO sign-off (Signed-off-by) as requested. Since the PR was closed due to inactivity, could you please reopen it or let me know if you would prefer me to open a new PR? Thanks!

umireon:
@henriquegeremia Hello, thanks for update! Unfortunately, this part was completely rewritten and this change is no longer valid. Of course, component-level comparison is implemented!
https://github.com/royshil/obs-backgroundremoval/blob/main/src/UpdateConfig/UpdateConfigSemver.cpp

PR #923, opened 2026-07-18
Fix RVM inference on GPU execution providers

### LINKED ISSUE #759: Windows and GPU version: Inference device only works on CPU, the other options only processes one frame and then stops
### Operating System

Windows 10

### Other OS (if not listed above)

_No response_

### Plugin Version

1.3.6

### OBS Studio Version

32.0.4 64-bit

### OBS Studio Log URL

https://obsproject.com/logs/kt0CsT9q2pFIVk4z

### OBS Studio Crash Log URL (if applicable)

_No response_

### Expected Behavior

the advance options let you change the inference device, which by default is CPU, and since I have a GPU and I installed the Windows and GPU version, I want better performance and results with the help of the GPU for removing background.

### Actual Behavior

My pc has a GPU, I installed the plugging version that supports GPUs, "inference device" is by default on CPU, when I change it to GPU, the filter only processes one frame and it stops removing the background, it keeps the only frame it removed the background, and it only works again when changing into CPU.

### Steps to Reproduce

activate "advance settings" in the filter options of this plug-in.
change to GPU - CUDA in "inference device".
the filter stops processing frames to remove BG and only works again when reverting to CPU.

### Additional Notes / Screenshots

Image

Image

### Confirmation

- [x] I have attached the required OBS log file(s).

burneikis:
# Fix RVM inference on GPU execution providers

Preface: This change was written with claude fable, I am not a C++ Systems developer. I have manually tested this change. I hope to get this issue fixed, and maybe this helps. I will not be offended if it's rejected.

To be clear, I have read contributing.md I understand to a very basic level whats happening here.
I am keeping this a draft while I try and understand this fully.

## Summary

Fixes #759 (RVM / GPU inference producing a single frame then freezing or crashing).

Two changes:

1. **Use the original `.onnx` model for RVM** instead of the pre-optimized
   `.with_runtime_opt.ort` variant, which is serialized with CPU-EP-specific
   optimizations and breaks GPU execution providers (CUDA, TensorRT, etc.).
2. **Let ORT allocate output tensors** in `runNetworkInference()` and copy
   results back into the pre-allocated CPU buffers with bounds checking.
   Pre-binding CPU output tensors misbehaves on GPU EPs for stateful
   multi-output models like RVM. (Same root cause #760 attempted; this goes
   further by also replacing the CPU-targeted `.ort` model.)

Notes for reviewers:
- CPU inference results are unchanged for all models. Overhead is one output
  tensor allocation + copy per frame (negligible). Output shape mismatches are
  now a logged warning with truncation instead of an ORT error.
- For RVM on CPU, session creation is slightly slower (graph optimization now
  happens at load instead of being pre-baked); inference speed is unchanged.
- Only RVM's model file changes; all other models keep their `.ort` files.
- `required_operators_and_types.with_runtime_opt.config` may need regenerating
  since the CI ORT build uses `--include_ops_by_config` (it is not a
  `--minimal_build`, so loading ONNX-format models is supported).

### Automated

- [x] CI builds pass on all platforms (Linux, Windows, macOS), including the
      custom reduced-ops ORT build loading the `.onnx` model.
- [x] clang-format check passes.
- [x] REUSE compliance check passes (new `.onnx.license` added, old `.ort`
      and license removed).

### Manual - GPU (the fix)

On a machine with an NVIDIA GPU (Linux CUDA build, and Windows GPU build if
available):

- [x] Add Background Removal filter, select RVM model, set inference device
      to CUDA: background is removed continuously (not frozen after 1 frame),
      no crash, no ORT errors in the OBS log.
- [ ] Repeat with TensorRT if available.
- [x] Toggle CPU -> CUDA -> CPU at runtime: filter keeps working after each
      switch.
- [x] Run for 10+ minutes to check stability and for memory growth
      (watch RSS / GPU memory).

### Manual - CPU

- [x] RVM on CPU still works (temporal consistency intact - matting should
      remain stable across frames, confirming recurrent state passing works).
- [ ] Compare CPU frame time before/after this change for RVM (expect no
      measurable regression from the added copy).
- [x] First filter load with RVM: slightly longer session creation is
      expected (graph optimization now happens at load); confirm it is
      acceptable (< ~1 s).

### Regression - other models

The inference change affects all models, so spot-check each on CPU:

- [x] SINet, MediaPipe, Selfie Segmentation, Selfie Multiclass, PPHumanSeg
      (segmentation models) produce correct masks.
- [ ] TBEFN / URetinex / SGLLIE (enhancement) and TCMonoDepth (depth) still
      produce correct output.
- [x] No new "Output size mismatch" warnings in the OBS log for any model.
- [x] Packaging: installed plugin data contains `rvm_mobilenetv3_fp32.onnx`
      and no stale `rvm_mobilenetv3_fp32.with_runtime_opt.ort`.

## Pull Request Checklist


umireon:
> This change was written with claude fable, I am not a C++ Systems developer.

Not acceptable at all. Closing. Please open an issue instead of PR.

umireon:
We accept Gen-AI based PRs only if we can assume the author has ability to verify their code safely. This is the meaning of the policy on CONTRIBUTING.md.

burneikis:
No problem, hope this code can still be used for reference / resolving the issue. 👍

umireon:
Thanks. We plan to remove the RVM model because it is too large and this problem will be removed.

burneikis:
> Thanks. We plan to remove the RVM model because it is too large and this problem will be removed.

Oh really? I've found this model to be the most useful, and personally, it's the only one I use, (with GPU, using my patch).

Is this decision final? Is there a better suggested option? I appreciate the development effort/cost required to maintain this sort of thing.

umireon:
@burneikis The computational load of RVM is too large for many users and they have reported issues frequently. The large footprint of RVM is related to its inference cost. Of course, I know the quality of RVM matte is quite good. Therefore, I'm considering generating fine mask from coarse matte like the approach implemented in [LBRL](https://github.com/kaito-tokyo/live-backgroundremoval-lite). If you want us to keep RVM within our plugin, please tell us why you need to use RVM and what is the key difference from every other model. We are always open but need evidences from users because we don't have telemetry in our plugin. Thank you for asking us about the current situation!

burneikis:
> @burneikis The computational load of RVM is too large for many users and they have reported issues frequently. The large footprint of RVM is related to its inference cost. Of course, I know the quality of RVM matte is quite good. Therefore, I'm considering generating fine mask from coarse matte like the approach implemented in [LBRL](https://github.com/kaito-tokyo/live-backgroundremoval-lite). If you want us to keep RVM within our plugin, please tell us why you need to use RVM and what is the key difference from every other model. We are always open but need evidences from users because we don't have telemetry in our plugin. Thank you for asking us about the current situation!

I will test this and put together a better explanation / reasoning and get back to you 👍

burneikis:
I did a quick test running the models side-by-side with my webcam to best showcase my preference.
this was run directly in Python, outside of OBS, RTX 4080, Logitech BRIO, 1280x720.

RVM GPU vs CPU:

- RVM on CUDA: 5.9 ms
- RVM on CPU: 20.4 ms

I also emulated the LBRL pipeline to test the coarse-matte idea (MediaPipe 256x144, fast guided filter, temporal smoothing, motion gate).

headshot

As you can see, the RVM is much more accurate / less transparent-spill

IMO, the key difference is the sharp/hard line

compare_hands

motion

So my request is to keep RVM, but only offer it when a GPU execution provider is active.
That being said, I understand the maintenance and implementation cost, so if you want to remove it, idm using my fork.
Thanks for your time and consideration.

umireon:
I agree with keeping RVM for large-sized portrait. The missing key from our plugin is clarifying the situation for which models are suited. Are you good with this conclusion?

burneikis:
> I agree with keeping RVM for large-sized portrait. The missing key from our plugin is clarifying the situation for which models are suited. Are you good with this conclusion?

Yeah, the main thing is that the model is slow (especially without GPU), I would only recommend for GPU use.

umireon:
I don't believe we can recommend RVM because GPU on many systems doesn't have enough resources for this model.

burneikis:
> I don't believe we can recommend RVM because GPU on many systems doesn't have enough resources for this model.

Oh I see, so even for average GPU it's still to much?
Could you detect available resources, or just put a noticeable disclaimer, requiring a strong GPU?

umireon:
Such an intelligent detection is so great for users. However, it is hell for developers because it enlarges our responsibilities to resolve errors on every specific equipment. So, my answer is no.

burneikis:
Fair. Perhaps just a disclaimer on the Model Selection, or in this project's README.

umireon:
My point of view is to add a model guide we have. I'd like to tell users that the only recommended model is MediaPipe and they can choose another model like RVM at their own based on the knowledge of models.

PR #948, opened 2026-08-13
Fix broken star history chart

Dessalines39394:
The star history chart in the README has stopped working, which leaves the project page without a usable overview of its growth. This change switches the chart to a working service so it renders correctly again.

This is a documentation-only change and has no effect on the plugin itself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,9 +120,9 @@ This video on YouTube will take you through the major parts of the code and expl
 </div>
 
 <picture>
-  <source media="(prefers-color-scheme: dark)" srcset="https://api.star-history.com/svg?repos=royshil/obs-backgroundremoval&type=Date&theme=dark" />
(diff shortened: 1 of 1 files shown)


royshil:
@umireon what do you think? it's fine I suppose

umireon:
I suppose we should remove the direct link to the image because I concern some privacy risks for users. Instead, we can add a static snapshot for star history. Is this fine?

Dessalines39394:
FYI Github proxies and caches all images, so that users will never query third-parties directly

umireon:
> FYI Github proxies and caches all images, so that users will never query third-parties directly

This statement is not like what we can accept. Since it originates from third-party, we cannot say that this is completely safe for users.

umireon:
We have decided to replace star history with a static image included in our repo. Thanks for opening PR, but we will go to the other way. Closing it with gratefulness ❤️

Dessalines39394:
Sure, that works. Thanks for the update

PR #970, opened 2026-08-28
Bump version to 1.5.0

umireon:
Bumping minor because the minimum requirement of OBS Studio is bumped to be 32.

Related to #970

## Overview

This pull request introduces semantic versioning (semver) parsing and comparison utilities to improve version handling throughout the project. It updates the version to 1.5.0 and replaces simple string comparison with robust semver logic for update checks. Additionally, it adds comprehensive unit tests for the new semver functionality.

**Versioning and Metadata Updates:**

* Bumped the project version from 1.4.1 to 1.5.0 in `VERSION`, `manifest.json`, `package.json`, `package-lock.json`, and `PLUGIN_VERSION` in `plugin-support.c`. [[1]](diffhunk://#diff-7b60b8e351cbb80c47459ffe2c79f1a26404871f49294780fe47ad0e58c09350L1-R1) [[2]](diffhunk://#diff-529de9be773dc420f66b65c56f42854e77e95a5e9859c3a5fce0707d0e028f03L4-R4) [[3]](diffhunk://#diff-74314f72f6616d400be4cb0d161867d8148f23c22d4c12c7188aedb129c00c59L4-R4) [[4]](diffhunk://#diff-8ae2b9f55773fd4d16657df5569a0d5acdbd86ed93c4fd26a67e77bced290d33L3-R9) [[5]](diffhunk://#diff-58fac19b479653af347e3608ccf20d120774e6058a794610d526cdc752936d72L13-R13)

**Semver Parsing and Comparison Utilities:**

* Added new functions in `UpdateConfig.hpp` and implemented them in `UpdateConfigSemver.cpp` for extracting major, minor, and patch components from a version string and for comparing two versions using semantic versioning rules. [[1]](diffhunk://#diff-d7ff9d99c7baca9cec906a78e10b96316036c72606d5370d1f9474348a784586R7-R21) [[2]](diffhunk://#diff-c6ed38343db4036fb68cdfe53a5e650785997a4c403c3b6f7d20ec7032c385a6R1-R61)
* Updated CMake build files to include the new `UpdateConfigSemver.cpp` source file.

**Update Checking Logic:**

[...]

PR #972, opened 2026-08-29
Assert the depth instead of the full model dtype

bxdxnn:
This fixes an assert crash if you have e.g. SINet selected as the model.

## Problem description

`cv::Mat::type()` contains the bit depth of each channel and the number of channels encoded within a single value. Possible values of `cv::Mat::type()` are `CV_32FC1`, `CV_32FC2`, and so on.
This assertion requires `srcMat` stored in Float32, and is not intended to restrict the number of channels.
`CV_MAT_DEPTH(dtype)` extracts only the number format and this will fix the root cause of crashes.
Note that the RelWithDebInfo configuration might disable any `assert()` macros and this is possibly why this bug is left untouched until now.

## Pull Request Checklist

- [x] I have included the required license headers.
- [x] I am confident with my code integrity.

### DIFF
--- a/src/models/Model.hpp
+++ b/src/models/Model.hpp
@@ -64,7 +64,7 @@ static void chw_to_hwc_32f(cv::InputArray src, cv::OutputArray dst)
 	const int height = srcMat.rows;
 	const int width = srcMat.cols;
 	const int dtype = srcMat.type();
-	assert(dtype == CV_32F);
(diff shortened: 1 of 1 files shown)


umireon (approved):
Valid fix. Thank you for great contribution!

umireon:
1.5.0 release will happen after this PR merged. #970 is just a bump PR.

umireon:
I will back in 1 hrs to release 1.5.0.

